PR #93, opened 2015-01-05
Fix dialog window focus

yazilliclick:
Fixes issues with dialogs not returning to focus when app is switched to and also showing up in alt+tab list separate from the main program.


EmmittJ:
Will look at this when I get home.

SpaceOgre:
Nice work.
Should add this to the update window as well :)

yazilliclick:
Thanks.  Made the same changes to all the message dialogs called from the main window.

SpaceOgre:
Looks good so will merge it now :)

Issue #70, opened 2014-12-10
Feature Request: Tree comparison


impacienco:
Not sure how easy this is to code (and let me know if you prefer that I leave requests on the forums/reddit instead of here).

It would be great to be able to select two trees/builds and get a one-click comparison that would show the net differences between the builds (e.g. +20% elemental damage, -40 dex).

Ttxman:
Actually implementing basic tree comparison was easy...
I'm sorry i had probably duplicated (or conflicted with) some of BlueManiacs code, but nothing was happening and i really need to make a new build :)

EmmittJ:
Alright, well it's no problem really. That's what happens in a project that
is open source. People develope things ;)

Version switching won't be implemented into the app. It's just too much to
keep working at a time. We can try and be retroactive but there are too
many trees to worry about, and people would expect it to work 100%

PR #94, opened 2015-01-05
Basic Tree comparison

Ttxman:
Selected build (in the saved builds expander) is highlighted (paths and passives) by the green color
Attributes in "PassiveAttributes" expander now shows difference to the selected build
  - attribute have bigger value than selected build: green colored "+number" at the end of attribute line
  - attribute have lesser value than selected build: red colored "-number" at the end of attribute line
  - attribute have same value as selected build: nothing :)

Passive attributes in the selected build that are missing from the current build are automatically added with value 0

poeskilltreecomparison

solution for issue #70, maybe :)

### DIFF
--- a/WPFSKillTree/SkillTreeFiles/Compute.cs
+++ b/WPFSKillTree/SkillTreeFiles/Compute.cs
@@ -2270,7 +2270,7 @@ namespace POESKillTree.SkillTreeFiles
             IsWieldingShield = MainHand.Is(WeaponType.Shield) || OffHand.Is(WeaponType.Shield);
             IsWieldingStaff = MainHand.Is(WeaponType.Staff);
 
-            Level = skillTree._level;
+            Level = skillTree.Level;
             if (Level < 1) Level = 1;
             else if (Level > 100) Level = 100;
 
@@ -2310,7 +2310,7 @@ namespace POESKillTree.SkillTreeFiles
 
             CoreAttributes();
 
-            Implicit = new AttributeSet(SkillTree.ImplicitAttributes(Global, skillTree.Level));
+            Implicit = new AttributeSet(SkillTree.ImplicitAttributes(Global, Level));
             Global.Add(Implicit);
 
             // Innate dual wielding bonuses.
--- a/WPFSKillTree/SkillTreeFiles/Compute.cs
+++ b/WPFSKillTree/SkillTreeFiles/Compute.cs
@@ -2310,7 +2310,7 @@ namespace POESKillTree.SkillTreeFiles
 
             CoreAttributes();
 
-            Implicit = new AttributeSet(skillTree.ImplicitAttributes(Global));
+            Implicit = new AttributeSet(SkillTree.ImplicitAttributes(Global, skillTree.Level));
             Global.Add(Implicit);
 
             // Innate dual wielding bonuses.
--- a/WPFSKillTree/Model/Options.cs
+++ b/WPFSKillTree/Model/Options.cs
@@ -1,11 +1,36 @@
-﻿namespace POESKillTree.Model
+﻿using System.ComponentModel;
+namespace POESKillTree.Model
 {
-    public class Options
+    public class Options : INotifyPropertyChanged
     {
         public string Theme { get; set; }
         public string Accent { get; set; } //Controlled by Menu Accent Headers
-        public bool AttributesBarOpened  { get; set; }
+        public bool AttributesBarOpened { get; set; }
         public bool BuildsBarOpened { get; set; }
+
+
+        private bool _TreeComparisonEnabled = false;
+        public bool TreeComparisonEnabled
+        {
+            get { return _TreeComparisonEnabled; }
+            set
+            {
+                _TreeComparisonEnabled = value;
+                OnPropertyChanged("TreeComparisonEnabled");
+            }
+        }
+
+        private int _SelectedBuildIndex = -1;
+
+        public int SelectedBuildIndex
+        {
+            get { return _SelectedBuildIndex; }
+            set 
+            { 
+                _SelectedBuildIndex = value;
+                OnPropertyChanged("SelectedBuildIndex");
+            }
+        }
 
         public Options()
         {
@@ -14,5 +39,13 @@
             AttributesBarOpened = false;
             BuildsBarOpened = false;
         }
+
+        private void OnPropertyChanged(string caller)
+        {
+            if (PropertyChanged != null)
+                PropertyChanged(this, new PropertyChangedEventArgs(caller));
+        }
+
+        public event PropertyChangedEventHandler PropertyChanged;
     }
 }
--- a/WPFSKillTree/Model/PersistentData.cs
+++ b/WPFSKillTree/Model/PersistentData.cs
@@ -4,10 +4,11 @@ using System.Linq;
 using System.Windows.Controls;
 using System.Xml.Serialization;
 using POESKillTree.ViewModels;
+using System.ComponentModel;
 
 namespace POESKillTree.Model
 {
-    public class PersistentData
+    public class PersistentData : INotifyPropertyChanged
     {
         public Options Options { get; set; }
         public PoEBuild CurrentBuild { get; set; }
@@ -43,6 +44,7 @@ namespace POESKillTree.Model
                 Builds = obj.Builds;
                 CurrentBuild = obj.CurrentBuild;
                 reader.Close();
+                OnPropertyChanged(null);
             }
         }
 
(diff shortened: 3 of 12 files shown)


EmmittJ:
I will go ahead and look at this when I get home but from what I see on my phone it looks good. Builds are compared between the current tree and highlighted build in the menu correct?

Ttxman:
yes they are... it seemed like best solution
The crude part is loading the tree nodes for comparison .. methods for extracting nodeID from poeurl and values from node id should be probably moved to static methods to avoid unnecessary creating whole new SkillTree. I probably look into that more but it seems that there are lot of properties that could (should?) be changed to static values or singletons

EmmittJ:
Not sure what you mean by static values. The tree should load itself dynamically no matter what we are doing to it. 

I think getting @l0g0sys and @SpaceOgre opinions would be a good idea for this merge

Ttxman:
Just that its superfluous (and little wasteful) to allocate lot of things in each instance of SkillTree, like:
        private readonly Dictionary<string, Asset> _assets = new Dictionary<string, Asset>();
        public List<string> AttributeTypes = new List<string>();
        public List<int> rootNodeList = new List<int>();
        public Dictionary<string, int> rootNodeClassDictionary = new Dictionary<string, int>();
        public Dictionary<int, int> startNodeDictionary = new Dictionary<int, int>();
        public HashSet<ushort> AvailNodes = new HashSet<ushort>();
        public Dictionary<string, float> BaseAttributes = new Dictionary<string, float>
        public Dictionary<string, string> NodeBackgrounds = new Dictionary<string, string>

All of these can be transformed to globally accessible (static access modifier) fields, which should reduce SkillTree creation time.

Every asset is loaded from hdd every time SkillTree is created and i am using SkillTree instance only to parse passive nodes from url.. Its just wrong :)

EmmittJ:
Was never meant to load more than one tree on the fly. You can go ahead and correct this if you would like :)

Ttxman:
If i start changing this it will be lot of changes ..

Ttxman:
Loading assests only 1 time per process really helps the UI speeeeed..

EmmittJ:
Nothing wrong with a lot of changes. If it is the correct way to do it then that is what needs to be done

WPFSKillTree/SkillTreeFiles/Compute.cs
```diff
@@ -2310,7 +2310,7 @@ public static void Initialize(SkillTree skillTree, ItemAttributes itemAttrs)
 
             CoreAttributes();
 
-            Implicit = new AttributeSet(skillTree.ImplicitAttributes(Global));
+            Implicit = new AttributeSet(SkillTree.ImplicitAttributes(Global, skillTree.Level));
```
l0g0sys: Use Level which is validated instead of skillTree.Level, e.g.:
Implicit = new AttributeSet(SkillTree.ImplicitAttributes(Global, Level));

l0g0sys:
Looks good, just need small fix according my diff comment :)

SpaceOgre:
I don't have access to a computer right now, but what I can see from my phone looks good. Some questions:
* The static objects is a good fix, I have been to lazy to do it myself ;), but how does it handle when the assets are re-downloaded from the menu? In this case at least every thing read from them needs to be read again since they could have changed.
* If the current tree is the same as the selected one, how does that work? Does it still compare everything or not since there is no need for it. 
* Is there some way to unselect the selected build, or some other way to hide the comparison? 

Like I said, can't test atm so some of these questions might be redundant.

WPFSKillTree/SkillTreeFiles/Compute.cs
Ttxman: OK. But in the original code, the property used was SkillTree._level, it worked exactly same way as now.

Will it be better to add bounds checking to property SkillTree.Level to make it work everywhere or change it just in Compute.cs ?

l0g0sys: Yes, you can check level bounds in SkillTree class, but still have to 
set Compute.Level property in Compute.Initialize method.

It looks silly to have code like this, it's matter of code consistency:
Compute::Initialize() {
...
Level = skilltree._level; // This is single point of Level intialization 
for all calculations within Compute class (no idea what is difference 
now between skillTree._level and SkillTree.Level).
...
skillTree.ImplicitAttributes(Global, skilltree.Level); // Here should be 
passed own Level property (no point reference external property twice, 
once own property has been initialized earlier).
...
}

Well, after second look it's even more mysterious why SkillTree method 
call has own property passed to :)
Is it really necessary to pass skilltree.Level to own method 
(ImplicitAttributes) of skillTree instance? :)

Ttxman: No its just lazyness to add nonstatic method in SkillTree that would call the static one :)

Ttxman:
@SpaceOgre 
- i forgot about this. But its simple to fix -> simpty set static bool _initilized to false and make new SkillTree
- yes it does, it highlights path, but differences with value 0 are are not shown. But its actually neat feature :), when you add new nodes, you will see difference between current build and the one you previously loaded.
- I don't think so.. maybe if you delete build or add new one ... but it will need small fix to hide the last comparison

WPFSKillTree/SkillTreeFiles/Compute.cs
Ttxman: some clarification
var st = new SkillTree()...;
st._level; //SHOULD BE private field, I will change it
st.Level;//public property to acess st._level
SkillTree.ImplicitAttributes(...);//static method to compute Health, mana gains etc. from int/dex/str and level

The actual change i made was: 
public ... st.ImplicitAttributes(attribs) method on SkillTree instance, that used only 1 nonstatic field - "sk.Level".
Is changed it to public static ImplicitAttributes(attrib, level) method to which you now have to pass the actual level.

The implicit attribute gains are same for all character classes and skilltrees, so i don't see reason to place the method that does not use anything (==pasive nodes), except level, from actual character as instance method.

l0g0sys: Ah, it's static. Didn't notice case change from skillTree to SkillTree in call to ImplicitAttributes.
Ok. It doesn't matter. SillTree.ImplicitAttribtues will be gone soon or later.
But for sake of correctness, you should pass Level (Compute.Level) to SkillTree.ImplicitAttributes() and not SkillTree.Level, or do level range validation on SkillTree.Level, otherwise calculations will be incorrect for level values outside of 1 ... 100 range.
Just change 2 lines in Compute.Initialize to:
...
Level = SkillTree.Level;
...
Implicit = new AttributeSet(SkillTree.ImplicitAttributes(Global, Level));

Ttxman: done

SpaceOgre:
I have now tried it out and have some more comments:
* Minion Instability and some other keystones does not show up in red when missing.
  * Note to rest: Conduit is under Charges and not Keystone
* The changes you made to the list have broken the word wrapping, it should look like this:
image
* The only way to deselect a build is to start drag an drop it, this have nothing to do with your implementation but it affect how it is used. So another way to deselect needs to be added and/or make it possible to toggle the comparison on/off. This could be added under the View menu and should be saved in persistent data, default should probably be off. (I would like to have the toggle either way :) )

Ttxman:
- The keystones are annoying .. the red coloring was based on absence of numbers. But the Minion instability have 33% inside text, so it was shown as "Minions .., dealing 0% .... [red]-33[/red]". I assume, that any keystone with number in text only had red negative number at the end.
- now all missing attributes are marked in red. so missing attributes are more distinguishable  than before
- wordwrap .. I have thought that they made textruns bindable in .NET4 but its not much better than before, so made it the old hacky way with converter. Whole text including the colored numbers is one textbox now so world wrapping works in all cases.

- i have added INotifyPropertyChanged interface to Options and PersistentData, to make it bindable from XAML
- Tree comparison can be disabled or enabled from View menu and with checkbox at the bottom of saved build list. And it is binded to Options and saved.
- and i have also binded selected speaker index to Options, so comparisons are now completely restored when application start
poetree2

EmmittJ:
This is looking really good man. Nice work! 
As for the side note, add this line: 
```
new[] {"Share Endurance, Frenzy and Power Charges with nearby party members", "Keystone"},
``` 
to GroupStringConverter.cs, line 18 before the other charges

Ttxman:
@EmmittJ i have added the line.

SpaceOgre:
Excellent job! Looks great :)
It seems that you know your way around wpf, which is nice for the project. Now I will have to go over some of my old code and fix some things.

Like you said the keystones are hard to compare correctly, like CI now gets +1/-1. But that is not so important, that is something we can look at later.

For me this can now be merged!

Side note for later: CI is not shown under the keystone group but under defenses.

EmmittJ:
Alright, go ahead and merge if you think it's ready :)

CI is under keystone because it's a keystone ;) can easily be changed
though

SpaceOgre:
Well atm it is not under Keystones ;)

EmmittJ:
Damn, ok I will do all that. Not sure how many I missed

PR #96, opened 2015-01-12
Improved node highlighting

MLanghof:
Hey, I made some changes that improve the way you can highlight nodes.
To this point you could highlight nodes by search or by right-clicking an attribute. I added the possibility to right-click nodes themselves to highlight them.
These highlights "stack" independently (I implemented somewhat rudimentary color coding).

In my eyes, this makes the "skill highlighted nodes" feature way more useful: You can search for "maximum life" and see all life nodes, then you manually (right-click) highlight all those that you want, then you clear the search box and click "skill highlighted nodes".

I've been looking into writing an optimizer for the "skill highlighted nodes" (the one you have currently implemented is just a greedy algorithm that will rarely ever find the most point-efficient tree). It's an NP-hard problem though (Steiner trees) and I've not quite found a decent optimization yet, but I wanted to get this preliminary work out there.


SpaceOgre:
Nice change, two things though:
1. Would be good if you could sync up with the master
2. There are no way to remove the highlights from right clicking an attribute, except to click another attribute. Some way to remove them is needed.
(3). Could be good to have a way to remove all highlights, maybe under **View**.

MLanghof:
Well this is weird... I don't recall adding the functionality to rightclick attributes to highlight them. Yet in the version i just downloaded to my PC that part doesn't work...

The functionality definitely did not directly come from me, see (here)[https://github.com/EmmittJ/PoESkillTree/blob/master/WPFSKillTree/Views/MainWindow.xaml.cs#L786]. It apparently broke in a recent release though (but it still works after the sync/merge I just did in my repo)...

Do you have a suggestion for 2.? I figured right clicking the skill tree background could work, but it's not really intuitive...

I just finished 1 (I think, please tell me if I made a newbie mistake).
I'll maybe look into 3. at some point.

EmmittJ:
1) I fixed yes :) since github gives the option to auto merge
2) maybe right clicking the node would remove it?
3) will be easy once 2 is done

MLanghof:
The issue is with right-clicking an attribute in the "Attributes" list. I just took it the way it was, and apparently it did not have a way to remove the highlight.

Skillnodes can already be un-highlighted properly by right-clicking them, that part works fine.

EmmittJ:
It un-highlights by moving the mouse out of the attributes list, which
honestly isn't the best solution.

MLanghof:
Huh? For me these things happen:
-Latest release ("Skill Tree Comparison", downloaded an hour ago): Nothing happens when right-clicking attributes.
-My local master: Highlight stays, mouse movement doesn't matter.
I guess the best way would be that another right-click on the attribute toggles that highlight back off.
Maybe the right-clicked attribute text should have its background slightly shaded so as to mark which one is producing the current highlight? The question would be what color to use (in order to work with the themes)...

EmmittJ:
You would want to use the ascent theme color :)

MLanghof:
Oh, I just saw #97 is related. That explains things.

EmmittJ:
Oh, yeah. It was broken in release, and I didn't catch it before pushing it.

SpaceOgre:
Will test again tomorrow, but by just looking at the code it looks good. 
About 3: I can add the menu connection if you create a RemoveAllHighlights method that I can call :)

MLanghof:
[There](https://github.com/MLanghof/PoESkillTree/commit/e8a93913516c1996dc79662e2cf4b399212eeb1f#diff-86591d87fc4bfd67edc41461789e4e03R792) you go!

SpaceOgre:
Did some testing and all my points are fixed, so as far as I'm concerned we can merge this. When that is done I will go ahead and connect the UnhighlightAllNodes to the GUI.

EmmittJ:
I don't see any problems. It looks good to me, so I would say merge it whenever you would like

MLanghof:
Oh right, I go by MauranKilom ingame, if you wouldn't mind changing that in the contributor list.

SpaceOgre:
No problem :)

Issue #98, opened 2015-01-13
"Skill highlighted nodes" is (too) greedy


MLanghof:
The current algorithm just greedily finds the closest target node and connects to it, then repeats. This way, it will never introduce Steiner points and will generally fail to find an optimal solution.

Example: I highlighted "Coldhearted Calculation" and "Void Barrier" (e.g. with regex "Void Barrier|Coldhearted") and have an empty Shadow tree. If I click on "Skill highlighted nodes", the current algorithm will first grab Coldhearted Calculation (7 points to grab vs 9 for void barrier) and then Void Barrier, resulting in [this tree](http://www.pathofexile.com/passive-skill-tree/AAAAAgYBRXwsRjy9UEKiLg5IXfLjhDY9bWzndNsaEHu-Og==) with 14 points.
[This tree](http://www.pathofexile.com/passive-skill-tree/AAAAAgYBDkhd8uOENj1tbOd02xoQe746D6uzDiHAsNg=) uses one point less though.

The problem of connecting a set of target nodes within a tree (with minimal point investment) is known as the [Steiner Tree problem](http://en.wikipedia.org/wiki/Steiner_tree_problem). It is NP-hard, so it will be quite difficult to devise an algorithm that finds the optimum reliably and within reasonable time for this kind of app.
I'm working on something for this, but I just wanted to create this issue so others who wonder if the current algorithm is optimal can learn about it. And possibly also to let others know someone is working on it.

EmmittJ:
Thanks, I was actually looking into it, but I am not very familiar with
algorithms. It would be great if you could fix it up :)

EmmittJ:
Hey MLanghof,
Just checking in to see how you are doing with the algorithm changes. If you need any help go ahead and let me know.
Thank you,
Emmitt

MLanghof:
I'm just quite buy with real life stuff atm. I got around to discussing my ideas with a friend and refined some stuff mentally, but I still have to code it. The coming weekend will likely see me getting on this.

EmmittJ:
Alright, thank you! Just wanted to make sure you were still here. I was
also discussing this problem with a friend, and going over the algorithm
you chose looks to be one of the better ways to do things. Just one
question: how well does Steiner's scale with hundreds of nodes? I haven't
done too much digging into it, but I didn't see a run time analysis on the
wiki page.

MLanghof:
[...]

PR #104, opened 2015-04-05
Implemented Steiner Tree optimizer for "Skill Highlighted Nodes"

MLanghof:
I'd consider the coding work for #98 done.

There's still the part of writing a decent documentation, but I don't feel like I'll get that done tonight. If you don't want to add that separately later, just leave this open I guess.

Oh right, and the window owner thing. Well, no rush.


EmmittJ:
If everyone wants to check and see if there is anything missing before merge that would be great
@l0g0sys @SpaceOgre

SpaceOgre:
Looks go to me, I will fix the window owner thing after the merge is done and other than that I think it looks good.

MLanghof:
Ok, code documentation has been pushed to my repo. If you ever wondered how I approached this, or why the number of used nodes in the best solution sometimes goes up, Steiner.cs is the place to look. :)

MLanghof:
Awesome! Now onto the next project...

PR #107, opened 2015-04-08
Aura calculator

nyashaint:
Hello guys,
This is aura calculator. Although it is not finished yet, basic functionality works.
It looks similar to https://poe.mikelat.com, with a bit of improvements.

What you can do with it:
1. calculate (surprise:)) mana/life reservation based on the current tree
2. modify life/mana/supported items
3. add/remove/minimize aura groups

image

Things to do:
1. add de-/serialization, so it can be saved with build
2. aura group names
3. aura effect calculations
4. visual mana/life reservation indication
5. support items from characters data


SpaceOgre:
Some quick comments, need to test more:
* Something is weird with the all the commits, but I guess that is some git quirk? (I'm not that git-savy so maybe it should look like this?)
* Looks and feels good, very intuitive
* Why all the Gem images? are they for something down the line?

Overall I think that this could be a good addition to the program as long as the connection to the tree is there. With that in mind I think that one of the most important features is to be able to use the current tree for the mana reduction/blood magic keystone settings. I feel like this plus saving should be in place before a release, since I think a lot of people will think that it will work like that.

What do the rest of you think?

EmmittJ:
Note on the git thing: The project was started in Dec so they had to merge the master whenever we updated something to keep current. 

I was going to ask the same question about gem images. The system works great and feels nice, but I would like to see tree interaction before merging into the main program. Something else that would be nice is instead of a 'popup' window make it so you can interact with both the tree and the menu. 

I am not sure exactly how you have it set up and that might take a bit, but I feel people would want live feed back on how their aura groups will change when they do something to the tree.

Other than that great work! It feels like it will be a great addition!

l0g0sys:
As standalone add-on to our app, why not. It's some extra functionality to offer to ppl, but i don't think there is any point in integrating it more with app.

In the end, integrated aura component of our app should just be list of auras, like some sidebar or something, where you can turn them on and off. But, as we don't have gem/gear swapping implemented yet, which would allow to change available auras, this tool is the only way to plan out your auras currently.

Looking at your TODO list, it should have separate element in PersistendData.xml to store those virtual aura configurations. It shouldn't go into build data at all.

Still, nice work! Think it will pass lot of time till this tool becomes obsolete :)

nyashaint:
I somehow overlooked this mess with the commits during pull request. This probably happened since I created this branch in December, then had no time to finish it, then played a bit with merge and rebases. If this will give you some problems with merge, I'll move changes to a new clean branch.
Gems images are there just for the style :) If you prefer to only have skill icons - it's easy to get rid on them.
I think selected auras are in a close relation with a build rather than with the general configuration, so the PoEBuild section in PersistendData.xml is the place where it should be stored.

SpaceOgre:
I totally think that the gems for the used skills should be there, but there are images for Exploding Arrow and other skills that are not in the calculator.

nyashaint:
I updated the UpdatedDB project so it fetches gem/skill icons for all gems we're using and adds a link to them to Items.xml.
I agree, although these icons might be used in future (for gem/gear swapping), means they could stay in source files, they are not needed in compiled binary.
I purpose to leave everything in UpdatedDB project, add a condition so names for unused icons aren't included in Items.xml, then remove unused icons from the resources.
What do you think about it?

l0g0sys:
Oh didn't notice icons being compiled in as resources. That's not good at all. Gems data and icons must be independent of binary completely.
  1. UpdateDB could just fetch gem/skill icons into Images/Gems and Images/Skills folders with simple name of gem plus extension. Should be cleaner than name_gem_icon and name_gem_skill_icon.
  2. Items.xml doesn't need to have links to those icons at all. Just waste of XML space due to icon file name being computable from gem name.
  3. I suggest to make Gem.GetGemIcon and Gem.GetSkillIcon methods which should return actual image. Case when there is no image should be handled too with some placeholder icon. There might be long delay between gem being released into game and having Wiki page created, so fresh gems can be added manually into ItemsLocal.xml and work with them asap.

nyashaint:
This solution is even better. Would do it like that.
As for the no image placeholder I want a white gem, but only found a socketed one
image

l0g0sys:
How is it going? :) Btw, can you make ZIP release pelase, so we can play with it without having to mess with another repository in VS? :)

SpaceOgre:
Just a nudge on this to see how it is going ;)

PR #109, opened 2015-04-10
Added a SkilltreeUndoState Class

EmmittJ:
I would like you guys to look over it and approve or not of it.
This is an update to a hack that I did for the undo feature.
As the title says I added a class which can be modified to hold anything and set anything we may need in the future. This is so we don't have to change undo logic, just the class. 

Tell me what you think and if it is good enough to be pulled.

### DIFF
--- a/UpdateDB/Properties/AssemblyInfo.cs
+++ b/UpdateDB/Properties/AssemblyInfo.cs
@@ -21,5 +21,5 @@ using System.Runtime.InteropServices;
 
 // Updated by release.xml. DO NOT MODIFY!
 [assembly: AssemblyVersion("2.0.0.0")]
-[assembly: AssemblyFileVersion("2.1.2.319")]
+[assembly: AssemblyFileVersion("2.1.2.320")]
 [assembly: AssemblyInformationalVersion("2.1.2")]
--- a/WPFSKillTree/Properties/AssemblyInfo.cs
+++ b/WPFSKillTree/Properties/AssemblyInfo.cs
@@ -42,5 +42,5 @@ using System.Windows;
 
 // Updated by release.xml. DO NOT MODIFY!
 [assembly: AssemblyVersion("2.0.0.0")]
-[assembly: AssemblyFileVersion("2.1.2.319")]
+[assembly: AssemblyFileVersion("2.1.2.320")]
 [assembly: AssemblyInformationalVersion("2.1.2")]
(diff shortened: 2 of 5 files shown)


SpaceOgre:
One thing: I'm not so sure that level should be part of the undo, since it does not affect the tree in any way. But I guess that depends on what we want undo to mean:
* Only tree changes
* All changes done to current build

If the first then I think we can remove the level thing and be happy. Otherwise more stuff needs to be added and more triggers for the undo stack needs to be added as well. Atm when you change level it will be added to the undo stack, it will only be added if you do a tree change as well.

l0g0sys:
Might as well have separate Undos. The one you made would be for "Undo tree change" and later we can add more complex build-wide Undo. So for start just tree undo will be fine.

nyashaint:
As we at that point, there is a bug where undo/redo doesn't update stats

EmmittJ:
Having just tree means that class doesn't actually properly change. The
current system is buggy, so I thought I would fix it up.

As for stat updates that is an easy fix. Just a function call. I can fix
that up when I get to a computer.

SpaceOgre:
Undo could call `LoadBuildFromUrl()` instead of 
```
Tree.LoadFromURL(tbSkillURL.Text);
tbUsedPoints.Text = "" + (Tree.SkilledNodes.Count - 1);
```
And if somethings are missing then it should be added there since it would be a problem when loading a build from url as well.

One thing I saw now is that tbUsedPoints is updated in UpdateAttributeList(), this feels wrong and it should be moved to somewhere else. I think class and used points should be updated in the same place when loading a tree.

l0g0sys:
At the moment UpdateAttributeList() is used as single-point of update 
for all tabs in contrast to its name, as there isn't dedicated UI update 
method yet.
So it's correct place for textbox to be updated there in current situation.

EmmittJ:
Basically the current implementation is working well and as it should, but
we aren't updating th UI off the current tree? Sounds like a better overall
fix than picking certain things to go back to.

l0g0sys:
If it helps, i don't care :) Actually, i don't have any idea what 
problem you trying to fix :)
Just stating the fact that in general it's better to have 1 place to 
update UI elements based on internal state, than having multiple methods 
messing with UI elements directly on multiple places.

EmmittJ:
Well, class and level aren't updated when you undo and redo, so you are
correct that it is because we aren't updating UI elements on load

l0g0sys:
And if those UI elements were updated using UpdateAttributeList() it 
would also solve:
     As we at that point, there is a bug where undo/redo doesn't update 
stats

PR #110, opened 2015-04-11
UpdateUI & CurrentClass Functions

EmmittJ:
UpdateUI is now a function and should be the one called when updating UI elements. Let me know if I missed a function

You can now get the current class from a loaded tree

### DIFF
--- a/WPFSKillTree/SkillTreeFiles/SkillTree.cs
+++ b/WPFSKillTree/SkillTreeFiles/SkillTree.cs
@@ -1036,5 +1036,33 @@ namespace POESKillTree.SkillTreeFiles
             }
             return false;
         }
+        public string CurrentClass()
+        {
+            string curClass = "";
+            var temp = new List<ushort>();
+
+            foreach (int node in _rootNodeList)
+            {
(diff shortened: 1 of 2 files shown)


EmmittJ:
What do you guys think @l0g0sys @SpaceOgre

l0g0sys:
Yep, makes more sense than UpdateAttributeList() :)

SpaceOgre:
I have not been at a computer all day but from what I see from the phone it
looks way better :)

EmmittJ:
Alright, I will merge it then!

SpaceOgre:
Not sure if completely correct line but around 1270 in main there is an unessesary setting of current class, since the call to UpdateUI comes right after.

EmmittJ:
Removed it.

SpaceOgre:
Oh and why do we need CurrentClass(), is not Tree.Chartype enough? Just asking, hard to get a good look on the phone :)

EmmittJ:
I actually just was looking at that and felt stupid lol It isn't needed either.

PR #115, opened 2015-04-24
Added support for optional Pre-release exclusion during check for update...

l0g0sys:
Added support for optional Pre-release exclusion during check for updates.
Added support for GitHub releases having multiple assets attached (one asset must have PoESkillTree in its name to be used for actual download).

Just some warm-up tiny work before all the upcoming mess lol :)
I found out, that i'm really rusty with C# after few months not working on project.

**Summary:**
Bit smarter lookup of releases on GitHub, specially with Regex matching of attached ZIP file in case multiple ZIPs are attached.
And CheckForUpdates has new Prerelease argument, which if set to true (default value is true for backward compatibility) it will also find Pre-release flagged release as candidate for update. If set to false, it will ignore all pre-releases.
I hope one day we will have Options UI to be able to set this, if one wants to use bit risky pre-releases.

### DIFF
--- a/WPFSKillTree/SkillTreeFiles/Updater.cs
+++ b/WPFSKillTree/SkillTreeFiles/Updater.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Text;
+using System.Text.RegularExpressions;
 using Raven.Json.Linq;
 using Ionic.Zip;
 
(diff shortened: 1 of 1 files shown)


EmmittJ:
Looks good to me. I don't see a reason not to pull. Some day we will have an options menu, but for now we are using the menu bar, which probably isn't the most elegant solution.

PR #126, opened 2015-05-26
UI localization

l0g0sys:
So, I'm pretty much done with localization for now.

Managed to localize almost whole application, except few parts which cannot be currently localized. I'll create separate issues which should be solved to have UI fully localized.

I wrote down also quite comprehensive document (Locale/README.txt) which describes all concepts and stuff, and should be read by developers and translators.

Except new stuff, there is important change, and that is, that PersistentData instance isn't “owned” by MainWindow anymore. New “owner” of PersistentData instance is App class, as it's required to have access to options before main window creation.

During localization, I also fixed some inconsistencies in terminology across the application, such as Build and Skill tree terms, which were interchanged in multiple places and didn't do well for localization. From now on, consistent terminology should be used, or translators will pull their hairs off :)

To try PoESkillTree in Slovak, download release from my repo https://github.com/l0g0sys/PoESkillTree/releases/download/2.1.4pre1/PoESkillTree-2.1.4pre1.zip.
And just add
```XML
<Language>sk</Language>
```
to Options element of PersistentData.xml, unless you have OS running with Slovak locale.


l0g0sys:
So what is the verdict on this one? Should i merge it?
I've got other stuff to pull and would like to close this one first.

EmmittJ:
Oh, yeah. Merge it! I thought you had already sorry

l0g0sys:
I reverted UI elements having issues when localized to rather use English and look good, than use localization and be distorted. Once the issues are resolved, it's piece of cake to localize them back.

Merge also won't affect anyone, except those running Slovak version of Windows. Those wil have it in Slovak language already.

EmmittJ:
Alright, sounds good!

EmmittJ:
Merged.

PR #132, opened 2015-06-03
Added application controlled pinning on taskbar

l0g0sys:
Added application controlled pinning on taskbar to avoid pinning of running shadow copy by OS.

This was longstanding issue with people pinning running shadow copy to taskbar instead of original executable by hand.

Found solution of using JumpTask to fake visual of pinning command and programatic way to perform "Pin to taskbar" operation, while having window flagged with PreventPinning, which disallows OS native "Pin this program blah blah" command.

All stuff is done in TaskbarHelper class, with almost no intrusion to rest of application.
Later it can be used for other fancy Windows 7+ stuff, like Recent builds jump lists, etc.

### DIFF
--- a/WPFSKillTree/Locale/Messages.pot
+++ b/WPFSKillTree/Locale/Messages.pot
@@ -478,6 +478,10 @@ msgstr ""
 msgid "Download completed or still in progress"
 msgstr ""
 
+#: Utils\TaskbarHelper.cs:79
+msgid "Pin this program to taskbar"
+msgstr ""
+
 #: ViewModels\PoEBuild.cs:30
 #, csharp-format
 msgid "{0}, {1} point used"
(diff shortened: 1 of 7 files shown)


EmmittJ:
Basically this just overrides Windows default pin?Would make sense since we don't want to be pinning the shadow copy. Everything looks good to me, maybe someone else has concerns. 

You can probably merge this :)

l0g0sys:
Well, in a sense yes. Tho, you cannot override OS controlled taskbar actions. So this just disables OS pinning and provides fake pinning action using JumpTask.
Also would be nice if someone could test it on Windows Vista or older which doesn't support pinning.

EmmittJ:
I don't have a machine that runs anything under Win 7.

EmmittJ:
Did you mean to merge and close or just close?

l0g0sys:
Lol, looks like i just closed it :)

EmmittJ:
You want it merged?

l0g0sys:
Just did it. Thanks:)

EmmittJ:
Email is a little behind :p

PR #140, opened 2015-06-19
Fixes for localiztion

### LINKED ISSUE #127: Menu Theme/Accent cannot be localized
Menu Theme/Accent cannot be localized due to use of MenuItem.Content directly. The C# backend should not use actual displayed name of theme colors and dark/light option when they are going to be translated.

### LINKED ISSUE #128: Class/Filter ComboBox in MainWindow cannot be localized
Class/Filter ComboBox in MainWindow cannot be localized due to use of ComboBoxItem.Content directly.
Pretty much same as MenuItem issue with theme colors.

EmmittJ:
Fixed #127
Fixed #128 

This should allow for proper localization of these items. @l0g0sys can you double check to make sure this will work?

### DIFF
--- a/WPFSKillTree/Views/MainWindow.xaml
+++ b/WPFSKillTree/Views/MainWindow.xaml
@@ -228,8 +228,8 @@
                     <MenuItem.Icon>
                         <ContentControl Template="{StaticResource ThemeIcon}" />
                     </MenuItem.Icon>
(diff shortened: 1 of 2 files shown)


l0g0sys:
Yep it works. If this Tag could be used with ComboBoxItem as well, it would also solve next issue with combobox'es like Class and Filter.

EmmittJ:
I solved that by using the name, like every other place in code. Content
is for display only and name should be the value associated with the combo

PR #141, opened 2015-06-22
Setup wizard  based installation/updating

l0g0sys:
I'm done with installer stuff too. Here are some notable changes.

## New features ##
  - dist-release.bat script now creates .exe package with setup wizard.
  - All required files to create setup wizard are in Tools\isetup folder. No need to install anything to make release.
  - Version.resx contains new strings which allows to make release without having to deal with Inno Setup script (release.iss).
  - Setup wizard asks for language to use during setup. Default Inno Setup language files have to be used (located in Tools\isetup\Languages), so those have to be managed separately from our localization.
  - Setup wizard displays LICENSE.txt file and asks for license acceptation before setup continues.
  - Setup wizard offers installation to Program Files by default, creates Start Menu (with uninstall link) and optional Desktop shortcut.
  - Setup wizard creates empty PeristentData file with language option set to chosen language during setup. This empty file is created by CreateSetupTemplate method in PersistenData class.
  - Setup wizard checks for running instance(s) of application and asks user to close this instance(s) before setup continues.
  - Setup wizard allows to install release alongside of older installed release based on Version.resx strings. AppId and ProductName strings have to be changed to achieve this.

## Changes ##
  - Removed shadow copying together with custom Taskbar pinning. Regular OS pinning will be used.
  - Application program files (executables and DLLs) are installed where selected during setup.
  - Application data (Data, Locale folders, Items.xml, PersistenData.xml files) are installed into roaming user profile (e.g. C:\Users\username\AppData\Roaming\PoESkillTree).
  - All application code accessing application data must use Utils.AppData API class.

## Game patch compatibility issues ##

As we know, at some point, it will be important to create release which must be installed alongside previous release.
Thus, allowing people to still use old release supporting 1.3 tree and calculations, together with new release supporting 2.0 tree and new calculations.
To manage compatibility of new release vs. installed release, 2 new Version.resx strings are used:
  - AppId: The application identifier is unique string which identifies application within Uninstall registry.
		   Thus, release with **same AppId** as existing installed release is treated as **update of already installed release** and doesn't create new Uninstall entry.
		   Changing this AppId will force setup to create new Uninstall entry during setup and treat installation as separate application.
           This string is also used as AppUserModelId to distinguish windows of application for purpose of grouping them on Taskbar.
  - ProductName: The ProductName is used to distinguish application release with friendly name.
                 The ProductName is offered as default folder name to be created in Program Files folder thus separating binary files from previous release.
				 Also it is used as Start Menu group name and Desktop shortcut name created during setup.
				 Finally, it is also used as name of folder within roaming profile to separate application data from previous release.

So, if current AppId is "{8B7A6A22-B6BB-40BB-856E-E0D1BB66DFB2}" and ProductName is "PoESkillTree" for PoE 1.3 compatible release.
The release compatible with PoE 2.0 will have new AppId generated and ProductName changed to "PoESkillTree 2" for example.
This will allow people to have "PoESkillTree" and "PoESkillTree 2" products installed at same time. Each with own application data, program files and Taskbar window grouping.
All further releases since then (when only ProductVersion is changed) will update only "PoESkillTree 2" application, until AppId and ProductName is changed again.

## Check for updates ##

  - New flag IsUpdate was added to Updater.Release class to indicate that release found on GitHub is an update of running installed application.
  - When installing an update, application will exit and run setup wizard in silent mode (no questions asked). Once update process finishes, the application will be launched automatically.
  - When installing a new product, setup wizard will be run in non-silent regular mode.


EmmittJ:
Instead of having more than one install directory, would it be possible that everything gets installed under PoESkilltree then having version numbers as subfolders? 
Ex: 
PoESkillTree
|____1.3
         |____App here
|____2.0
         |____App here

l0g0sys:
Maybe it could be hacked to work like that somehow.
My brief estimate is month or two to solve all the issues resulting from 
that model (access rights, fixed installation directories, uninstall 
logs, shared folder ownership, ...).
Any reason why that mess would be better than standard way (i.e. every 
other application is installed)?
At the moment, using standard model is breeze with Inno Setup.
Unless there is very good reason to not use standard way, i'd like to 
spend my time on something more useful :)

EmmittJ:
I completely understand, but other programs also don't keep archived
versions of their tools laying around for people to install, and if they do
(ex: adobe) they throw them all into one folder as separate programs.

l0g0sys:
Yeah, it might be nice to do that publisher named shared folder and stuff. It's used by publishers which have many many tools (notice only some publishers use that model).
But it's too much of work for just such cosmetic difference, which normal people won't notice, if they even care about all it.
In the end with setup wizard:
  1. You don't have to look for application, where it is installed at all.
      You have Start Menu, Desktop icon, taskbar pinning to launch it.
  2. Uninstall is automated so you don't have to care what to delete from where at all. Go to Uninstall program Control panel or use Uninstall Start Menu entry, click and done.

EmmittJ:
I see your point :) anyway this is pretty sweet!

l0g0sys:
Not to mention that it's nice that people can choose where they want to 
install it and how they want to name it.
Now i can install old version to disk C under "Awesome PoE tool but old" 
and new version to disk D under "Even more awesome PoE tool" :))
And tbh, once you are done with 1.3 to 2.0 transition lot of people will 
uninstall 1.3 as there won't be any point using it anymore and you end 
up again with single application.

EmmittJ:
Very true. I kind of forgot that you can customize the install location >.>

l0g0sys:
What i would rather spend my time on, setup wizard related, would be to 
provide way to install application into portable form.
@MLanghof mentioned that he likes portable apps, and i actually stumbled 
upon Inno Setup having support for such installation mode.
So there could be checkbox during setup like "Install application to 
portable device", so you could install it onto USB key for example and 
it would run purely from that location and you could take it anywhere.

EmmittJ:
I also like that it is portable. I just have it on a Google drive so all my
stuff is the same across all comps

l0g0sys:
So, it wasn't that much trouble and i added option to choose portable installation for example to USB stick.
Hope the description is clear about what's gonna happen when one chooses portable installation.

portable

EmmittJ:
Perfectly clear.

l0g0sys:
Can someone check the Wiki page "How To: Release" please ?
I've added new stuff there and want to be sure it's understandable and there are not any strange language/sentence constructs :)

SpaceOgre:
The updated wiki looks fine to me, easy to understand.

MLanghof:
Very impressive work, it looks very good!

PR #143, opened 2015-06-25
Added missing localization of messages.

l0g0sys:
Added Popup class to simplify generic error and information reporting.
Also added more detailed information for Check for Update to clarify whether found release is an update or new product release.


SpaceOgre:
Looks good.

Issue #114, opened 2015-04-23
Full Tree Optimizer/Skill Highlighted Nodes not getting ideal path


teuneboon:
I was playing with the "Skill Highlighted Nodes" function, it didn't get the ideal path for me:
![Example 1](http://i.imgur.com/IbfYwrL.jpg)
Why did it go through those 2 minion nodes while it could also have gone through just 1 crit node?

Full tree:
![Example 2](http://i.imgur.com/LTxpLtr.jpg)
Already fixed the mistake in this screenshot

MLanghof:
Thanks for this info. To be open about it, the optimizer is not technically _guaranteed_ to find the best solution, only *very likely* (because the task is "very difficult" from a mathematical perspective, *guaranteeing* an optimal solution would take a very long time).

I'll check out your example and see what I can do to tweak things.

Thanks for the report!

MLanghof:
After looking into this, you seem to have gotten somewhat unlucky (with your configuration, I got that result only very few times upon letting it run multiple times), but I opted to tweak something so the optimizer should be more robust (and yeah, this means it'll take longer, tradeoffs everywhere).

Thanks again for reporting this! I tried hard to come up with "difficult" inputs before, but there's always new ones that I need to watch out for. Please do inform me if you get a bad result again!

EmmittJ:
I can close this if you wish. It seems like you may have fixed the problem?

EmmittJ:
I guess 'problem' it will never be 100% :p

MLanghof:
Indeed, there's only so many different trees I can check in a given amount of time. And lots of common inputs have many orders of magnitude more possible tree solution candidates than that.

I can influence the balance between "can accept a worse candidate in order to move away from an *almost* optimal solution" and "pursue an almost optimal solution and optimize it as far as possible", but ultimately there's always "easy" and "hard" inputs.

You can close this issue if you want, but I'll still have to push the tweak. I guess I need to open a pull request every time for that?

EmmittJ:
Correct. Just like everyone else :)

PR #144, opened 2015-06-25
Steiner/Skill highlighted nodes improvements

brather1ng:
I made some changes to the SteinerTree classes that lead to more robustness (all trees I tested nearly always produce the optimal solution) and less runtime (3-6 times faster for normal sized trees).

What I mainly did are two things:
* The fitness function is now accurate. The Dijkstra algorithm was not far away from calculating shortest paths besides shortest distances so I added that functionality and used it in MinimalSpanningTree.UsedNodeCount.
* The loops that spawn individuals in GeneticAlgorithm are parallelized. This is a around factor 3 runtime improvement on my 4 core machine.

I also made some smaller tweaks that overall improve the genetic algorithm:
* no reduction of mutation.Age
* balance between populationSize and maxGeneration shifted towards populationSize
* higher maxGeneration for lower dimensions
* removed annealing from GeneticAlgorithm

I uploaded the test documentation I did: https://www.dropbox.com/s/lwva3ssmjl5t8ue/SkillTreeSteiner.ods?dl=0

Also as a side note: I don't know if it is only for me, but if you first checkout this project, git says a couple of files are modified because they seem to be uploaded with the wrong line encoding (crlf). Commenting out "* text=auto", saving and reverting the comment in .gitattributes fixes it. But it is still somewhat annoying.


l0g0sys:
Can't comment regarding algorithm changes, it's many years since i left university :p 
So i hope @MLanghof will look into it. He's our fresh math geek atm :)

Regarding .gitattributes.
The project is one mess regarding the line endings, that's why I introduced .gitattributes to normalize all line endings. The problem was addressed bit late, but it had to be done. I'm sorry for you trouble.

Guess, i'll have to revisit rest of the files, which are still not normalized and re-push them with correct line endings to avoid this issue in the future. Expect this weekend lot of changes ;)

brather1ng:
For the line endings it might be enough to do a fresh checkout and just commiting what git says is modified. 
But I'll open an issue with the files that seem to be incorrect later today.

l0g0sys:
No worries :) i'll find them, they can't hide :)

MLanghof:
Nice, I'm glad someone actually picked up on this! Will definitely check it out :)

MLanghof:
Looks nice!
I was afraid of storing paths in my implementation since storing O(n²) distances already seemed quite a bit (and I thought storing only one of multiple shortest paths doesn't reduce ambiguity, although I see now how the overestimation of costs really hurts), but it seems it's working out.

Since you also seem to have run a lot of practical tests successfully (and since we don't lose previous versions thanks to git), I'd suggest simply removing all the things you commented out (presumably for testing purposes) for keeping the code clean.

And yeah, line endings are currently really killing my Git diffs, it's kinda painful. But I don't know exactly what to do, will happily follow instructions though.

Once this is done you can probably close #114.

brather1ng:
Thanks!
I did think about that too, but since what is stored (small sets of shorts) with like 8000 distances calculated it doesn't turn out to be any significant amount.

Yep, was going to do that and correcting the commentation once someone checked it all.

I don't know if you have the same problem as me, but for me it was enough to go to `.gitattributes`, comment out `* text=auto`, save and give git some time (a few seconds) and then revert the change.
Doing that once for every fresh checkout/branch switch solved the problem for me.
The only problem is one file that is only partly wrong, for that a random mix of parts of the above and discarding the file seems to fix it. This needs to be fixed only once per clone though, for whatever reason.

brather1ng:
Cleaned up the documentation. It felt somewhat bad removing all the simulated annealing stuff and long explanation of the inaccurate fitness function. :)

From my part, this is finished now, if noone finds any bugs or stuff like that.

MLanghof:
I like how you kept the temperature which is actually part of the annealing (as is `acceptNewState()`) :D
It no longer decreases though, so you might want to change the wording in line 45. At this point it's effectively a hardcoded magic number that influences how likely it is for mutations to inferior states to be accepted (again, technically a concept of SA). If your tests including it were satisfactory it could be kept in one way or another (in the end it's a convergence speed accelerator), but the temperature part should probably be made less... Well, at least it should be put into the appropriate scope.

brather1ng:
Most of my knowledge about genetic algorithms is based on a few Wikipedia articles and reading your code, so yes, what I did is probably not technically correct. :D
A chance of accepting inferior states definetely proved to be helpful though. From what I read something like this is normally part of the selection process that in your code is just selecting the best 50% and mutation normally is also done different. So in the end `acceptNewState` is just a varation that has a good effect and happens to be annealing related, I guess.

I didn't test around much with the temperature itself but I think the whole algorithm is now in an acceptable state, so "correctly" removing annealing doesn't seem too appealing to me. :)
I get that the name "temperature" and `acceptNewState` being in a region called "simulated annealing" could be misleading though, so what would you call an appropriate scope for the `acceptNewState`-method? (I removed the temperature variable now and hardcoded it into the function)

MLanghof:
> A chance of accepting inferior states definetely proved to be helpful though
Of course! The non-standard thing (for a GA) here is the chance to *reject* inferior states (before the next generation even starts). It makes a lot of sense though, since there generally are a lot of potential steiner nodes that are just far off and will never contribute to a better solution, so doing this "sanity" check helps keeping the dna quality in the pool high.

You could just remove the region thing. Maybe also make a small comment how the formula is borrowed from SA and why it's useful here. `acceptNewState()` could be changed to something like "infantDeath()" or rather (how it's used now) "surviveChildhood()" but I feel it's a little morbid. The name is probably fine as it is.
As for the temperature: Its value could probably be played around with to gauge the effect it has, but ultimately it's a tradeoff between how aggressively the solution space is explored and how fast the algorithm "converges" (in a sense). If it works properly right now it would likely be fine to leave it untouched though.

brather1ng:
Thanks for that explanation, I added parts of it into the documentation so it should no longer be a confusing use of annealing terms.

EmmittJ:
If someone gets a chance to look at the conflicts before I get back to a pc that would be great :P

brather1ng:
The conflicts probably came from my other pull request. I'll solve them once I'm home.

EmmittJ:
Oh, whoops. Sounds good

brather1ng:
Conflicts resolved!

Issue #145, opened 2015-06-26
Make Skill tree optimizer to ignore (not select) certain highlighted nodes


l0g0sys:
Interesting idea at forums :)

https://www.pathofexile.com/forum/view-thread/996805/page/22#p10858775
>Do you think it would be possible to add the option to ignore selected nodes (maybe with middle mouse button) , so if i never want the automatic planner to pick certain nodes then it would try to find another way or if there is no way then it would give an error message?

Something like tri-state node highlightning? Non-highlighted, highlighted to be selected, highlighted to not be selected (different outline color/shape?).
And optimizer would be forced to not pick "don't select" nodes.

brather1ng:
Sounds like it could be useful. 
It might even be pretty easy to implement on the optimizer side (if just changing the scoring function to give bad scores when ignored nodes are included is enough for acceptable results).

For the node highlighting, I have absolutely no idea what is possible, but something like in the screenshot should look intuitive enough.
![crossed](https://cloud.githubusercontent.com/assets/7596346/8376193/9034026e-1c04-11e5-89d4-d267bee844ef.png)

brather1ng:
I have implementend this now locally. Both the optimizer and the GetShortestPath-thing don't select nodes you have middle-clicked. 
Should I make a new branch for it or is it ok to add it to my existing pull request?
![implementation](https://cloud.githubusercontent.com/assets/7596346/8392083/ceb2a50c-1cdd-11e5-8773-124c99d6c241.png)

MLanghof:
Well, this definitely puts more emphasis on handling unconnected graphs correctly. We need appropriate error messages since this couldn't happen before (although I did make a unit test that would check for it).

brather1ng:
Yeah, I somewhat handled that in my code.
Pushed my implementation as a branch to my fork now, see c40e9d623cde03534dae87a57e0aa4c1020db397
Going to open a pull request for it now hoping it doesn't conflict too much with my other pull request :)

MLanghof:
Makes me realize again that the Surface has no good way to MMB, so can't try it out right now.

Just out of curiosity, where in Germany are you from?

brather1ng:
Didn't think about MMB being accesible or not, kinda just copied the suggestion. If there at some point will be proper touch support for this program the key for this should probably be changed.

To your curiosity: I'm from Bremerhaven, going to university in Bremen.

MLanghof:
[...]

PR #146, opened 2015-06-27
Added Crossed HighlightState that is ignored by the skill tree optimizer

brather1ng:
Added another HighlightState for nodes that is accessible with shift+right-click.
Crossed nodes get ignored by the optimizer and SkillTree.GetShortestPathTo.
If omitting nodes disconnect the tree for the optimizer, a popup is shown (Visual Studio stops after the exception that gets handled later, so to see the popup, press F5).
See Issue #145


brather1ng:
From my part, this is finished now.

The new messages need to be added to the localization-thing, but I won't read myself into that since it will be done way faster by whoever coded that.
They are at the bottom of `HotkeysWindow.xaml` and in `OptimizerControllerWindow.xaml.cs`.

EmmittJ:
Looks good to me! @l0g0sys should be able to help out with the localization when I merge this :)

PR #153, opened 2015-07-11
Separated node tagging from highlighting

### LINKED ISSUE #149: Skill node highlighting ambiguity
There is ambiguity in node highlighting when right-clicking the attribute name in Attributes sidebar and right-clicking a node in Skill tree.

It is confusing to have green and blue circle highlighting with same meaning, but different way of access. Not to mention that you can have green circle with red cross together on same node.

Is it possible to have just one type of highlighting for node inclusion?

Using only green circle (easier to spot than blue one) should be sufficient.
For attribute name clicking to act as "all or nothing" toggle as it's current behaviour, the check for existance of corresponding highlighted node should be done first.
If such node exists, then unhighlight all highlighted corresponding nodes.
If no such node exists, then highlight all corresponding nodes.

brather1ng:
Fixes #149.

I don't know if "Check-tag highlighted Nodes" and "Cross-tag highlighted Nodes" are particularly useful.
If you don't think they clutter the menu, someone would need to add icons for the menu entries if we want to have icons for every menu entry (that might not be the case).

### DIFF
--- a/UnitTests/TestGeneticAlgorithmEasy.cs
+++ b/UnitTests/TestGeneticAlgorithmEasy.cs
@@ -32,11 +32,11 @@ namespace UnitTests
             SkillNode coldhearted = SkillTree.Skillnodes.Values.Where(n => n.Name == "Coldhearted Calculation").First();
             SkillNode voidBarrier = SkillTree.Skillnodes.Values.Where(n => n.Name == "Void Barrier").First();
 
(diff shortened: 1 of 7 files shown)


EmmittJ:
Should we get this in before I release?

brather1ng:
If you don't insist on having menu icons, I'd say yes, since it's finished in that case.

EmmittJ:
Menu icons don't really bug me so I will merge

PR #160, opened 2015-07-12
Updated Frenzy charge attributes

andrew-gardener:
Hi, 

I noticed that the 4% more damage frenzy charge attribute was missing and that the cast and attack speed attributes were the old values.

I went ahead and changed them (hope you don't mind, I'm new to this pull-request thing)


EmmittJ:
Good stuff, thank you!

PR #162, opened 2015-07-12
Fix for problem with Skill tagged nodes and small search spaces

brather1ng:
Fixed the problem described here https://www.reddit.com/r/pathofexile/comments/3cyk54/poeskilltree_220_the_awakening_update/ct0iiaf
Small search spaces took to long and directly adjacent tagged nodes let to a crash.
(also changed Console.WriteLine to Debug.WriteLine because I stumbled upon that)


MLanghof:
Damn, I should've finished the unit tests... They operate on small scales (at least the ones I made) and would've detected that...

PR #167, opened 2015-07-14
Fixed two more README.md typos

honzi:
Hello again, few more typo fixes.

* I did some research and I think `PoEURL` is actually correct based on the capitalization of the `<title>` of their website, so I changed them to that.
* I missed a missing `r` too.


EmmittJ:
Thanks again :)

Issue #183, opened 2015-07-19
Highlight similar skill nodes feature


brather1ng:
Would work basically the same way as the feature of the official skill tree (https://www.pathofexile.com/passive-skill-tree/, checkbox in the bottom left): all nodes having (or containing) the same set of attributes as the hovered node get highlighted.
Maybe not as a checkbox, but on key press. For example shift + hover over a skill node highlights similar skill nodes.

Do you guys think this would be useful?

SpaceOgre:
Yeah it might be useful, but the way it is implemented in the official tree is to strict. If I hover on 10% increased fire damage then Fire Walker will not be highlighted.

brather1ng:
The official tree only highlights other nodes with the same name, so yeah, that is to strict.
The question is, do we want to highlight other nodes that have *all* the attributes (or more) the hovered node has or nodes that have at least *one* of the attributes.

EmmittJ:
I don't mean to be that guy but isn't that was the search bar is for?

brather1ng:
Yes, this is basically just a (for the user much faster) alternative to the search bar.
That is why I asked for opinions :)

EmmittJ:
It might be good, but what do you do for double+ stat nodes? Do you
highlight everything that contains all of the stats of nodes that contain
at least one?

brather1ng:
Do you mean the same thing I wrote above?
>The question is, do we want to highlight other nodes that have all the attributes (or more) the hovered node has or nodes that have at least one of the attributes.

If yes, well, I don't know. We could test both and take what feels better.

EmmittJ:
Dammit, yes. I missed that email lol like you said we can try it out and go
from there

PR #190, opened 2015-07-21
Added highlight similar nodes feature

brather1ng:
Holding shift while hovering over a node in the skill tree highlights all nodes that contain *all* (can easily be changed to *any*) of the attributes the hovered node has.
*All* feels better imo.
See #183.


SpaceOgre:
Works like a charm. A small thing though: I think that it would be good to hide the tooltip for the duration of the shift press, as it can hide similar nodes.

brather1ng:
Yes, that is better. Added tooltip hiding.

Issue #194, opened 2015-07-23
Update progressbar says "Downloading skill tree assets"


brather1ng:
If you update the program with "Check for updates", the progress bar says "Downloading skill tree assets".

EmmittJ:
Same for if you re-download item assests ;)

We use the same loading window for all download processes

Ttxman:
shouldn't we change it into "Downloading POE skill tree assets" ?

l0g0sys:
Nah, DownloadItemsWindow.xaml should be refactored to DownloadWindow with some property, where you can set text to display. That way it can be used to download anything. And when used with Updater, just set property to "Downloading installation package" or something, while when being used with SkillTree, property will be set to it's current text "Downloading skill tree assets".

Issue #174, opened 2015-07-15
Remember window position and size (and fullscreen state)


Tvangeste:
It would be nice for the PoESkillTree planner to remember its latest window size and position, as well as fullscreen state.

Otherwise, user have to rearrange the window every time he/she starts the planner, over and over again.

Me, I would be happy if the tool could at least remember the fullscreen state. :)

SpaceOgre:
Yes this would be nice.

Implementation tips: http://stackoverflow.com/questions/847752/net-wpf-remember-window-size-between-sessions

Issue #150, opened 2015-07-10
Change "Load Build" to asynchronous


SpaceOgre:
Same as in #77 but for load build instead. This mostly affect poeurl links but, could just as well do it for all. Maybe even show a loading icon somewhere so that the users knows that the program is doing something.

PR #197, opened 2015-07-30
Multiple smaller changes (LoadingWindow text, saving window position, asynchronous tasks)

brather1ng:
Fixed some smaller issues:
- #194: LoadingWindow has the info text as a property that can be set over StartLoadingWindow.
- #174: Mahapps has a property `SaveWindowPosition` that does all the work.
- Added a clear button for the search text field (a property from Mahapps I stumbled upon, seems useful in this case).
- #150: Running background tasks are shown in the title bar. Implemented checking for updates and resolving poeurl/tinyurl links as such tasks. Below is a screenshot of how it looks, do you guys think that is a proper way of showing that work is happening?
asynctasks  
Also, is there any particular reason the app doesn't use .net4.5? Because the async/await stuff would make this a lot easier to implement.


EmmittJ:
I am pretty sure at some point this project was in .net 4.5, but I am not sure what happened there. 

Anyway this looks really nicely handled for the background stuff

SpaceOgre:
I don't think there should be any problems with going for .net 4.5.
And I agree with @EmmittJ that the background solution is really nice.

About the ```SaveWindowPosition``` from what I can tell it will store this in AppData somewhere, which is fine but if you have the program installed as portal it might not be so nice. But that is up for discussion of course.

l0g0sys:
So maybe using Registry for storing UI state?

brather1ng:
>About the SaveWindowPosition from what I can tell it will store this in AppData somewhere, which is fine but if you have the program installed as portal it might not be so nice. But that is up for discussion of course.

Yes, it stores the user.config in AppData by default. For portable it should probably store it in the program folder (so Registry wouldn't be better imo).
I'll try to produce an installer and check where it saves it with a portable installation later. Maybe it already does it correctly, else I'd need to find a way to modify the user.config save path or a completely different way of storing the UI state.

SpaceOgre:
Yes, storing it in program folder for portable would be best.

brather1ng:
This is done now (they are saved in AppData.GetFolder("Settings"), but needed an update of MahApps.Metro.
I also upgraded to .Net 4.5 to use async/await for the background task stuff.

So the commit contains a ton of changes, the relevant changes are in Settings.cs, CustomSettingsProvider.cs and MainWindow.xaml.cs.

I'll try to find out if the upgrade has broken anything later.

PR #208, opened 2015-09-09
Fix for item import

Yskuma:
Fixed this for my own use so I thought I'd create a pull request to save somebody having to track down the bug.

Based on the fix Procurement did here: https://github.com/Stickymaddness/Procurement/commit/f8fff0c6d573ed15075dd860415bab95ca1c8f4f

You may want to handle it another way but it seems to work.


brather1ng:
Imported items look pretty ugly with those weird tags in their name , so I guess it would be useful to merge this.

Issue #185, opened 2015-07-20
New features 


dekelpilli:
Base on: https://www.reddit.com/r/pathofexile/comments/3dqqov/poemate_new_iosandroid_skilltree_app_full_jewels/

I would very much like the feature where you can see the total increased damage on your main skill (so, say you're using Flameblast and you have 12% area damage, 40% fire damage, 16% elemental damage with spells and 60% spell damage, there will be something that says "Total increased Flameblast damage: 128%, or broken down into the different damage types as shown in the video). The same thing for attack/cast speed and other skill-relevant modifiers (such as showing a total crit chance increase for a spell, totalling both global and spell crit) Being able to see how much life and mana you'd have with 'x' given life/mana/strength/int on tree would also be very useful. Integration of interaction with unique jewels would also be cool, but at a lower priority.

Issue #106, opened 2015-04-08
Full tree optimizer


MLanghof:
The original reason why I started looking into this project and worked on #98 was the vision of a tool that builds an entire tree based on specified user criteria.

There's so many tradeoffs (life, mana, damage, cast speed, mitigation etc.) that need to be made when devising a tree, and so many different nodes and node values (with widely varying point investments) that it's hard for a player to create something optimized. Only very few people actually do/check the math. For example, it's generally worth to take the melding cluster (25% ES) over spending those four points in the scion ES wheel (24% after you got Foresight), even if you only care about ES.
Or another example, I often wonder whether I should get more cast speed or rather more spell damage in my tree at a particular point in order to optimize my dps. That obviously depends on my gear and my current tree, but that's data that can already be fed to the existing skill tree tool.

It's only a matter of letting the user communicate his desires and priorities about the tree. Well, "only" doesn't really nail it, the required UI work (and thought) is the entire challenge here.
My current idea is that users can select a node (control-click? Or go into a special mode and right click?) and then choose how much he values certain attributes that node has (which will "apply" to all nodes with those attributes).
Going one step further, either the user should be able to combine attributes (dpsMulti = (1+spelldmg)*(1+castspeed)) or there should premade functions that compute things of interest (total life, armor...). In particular, Compute.cs (I was really impressed somebody sat down and basically recreated the engine!) will provide accurate calculations for whatever setups and gem links people want.
Then there needs to be a way for the user to weight those aspects in a certain way. Here my idea is that the user selects a desired target value and how important it is to reach that value. A Mjölner char has pretty set requirements for str and int (so that would have full priority), but maybe setting desired life to 160% increased and giving it lower priority could give you 150% life with much better dps. You get the point.

Interestingly enough, all the "hard" algorithmic work is already completed with #98: Instead of searching for the set of steiner points with the lowest point investment, all nodes possibly affecting the user defined tree value will be part of the searched node combinations. The genetic algorithm doesn't really care, it only needs a fitness function (providing which will be the challenge). It's obviously going to take longer to find the optimum, but that's a given.

I still don't really know how to represent "value" for this, and how the individual aspects should be combined (multiplied? added?). And especially how to make this somewhat transparent for a user.

So these are the questions I'm asking myself at the moment:
-In what way should the scores for e.g. life and dps be defined, and how should those be combined into a final score/(fitness) value?
[...]

PR #210, opened 2015-10-05
Skill tree generator - first set of features

brather1ng:
The skill tree generator implementation (see #106) is now at a point where it has enough features to be worth including in the program in my opinion.

### New features

The user can input basic attributes ('#% increased maximum Life', '#% increased Attack Speed with Bows', everything that appears on skill nodes) and pseudo attributes (where the basic attributes are simply added together given settings for weapon class, tags and off hand style, like '#% increased (total) Physical Damage') as constraints with a target value and a weight for which the program tries to find the optimal tree with respect to those constraints and a given number of points available. What is not yet implemented are constraints that include the gear in any shape or form and an automated solver that derives the constraints in some kind of heuristical way.

### UI overview

You can finde the generator under the menu at "Tools -> Nodes -> Skill Tree Generator". In the window are some general settings and a tab for every different solver. "Advanced" contains the new functionality, "Automated" does nothing atm and "Tagged Nodes" is exactly the "Skill Tagged Nodes" feature (which is multiple times faster now), included there for testing purposes.

In the "Advanced" tab, "Attributes" is for the raw attributes, "Pseudo Attributes" is for combined raw attributes, "Combined Attributes" will later be for stuff including the gear and "Settings" contains settings related to which raw attributes are calculated into pseudo attributes atm. One thing worth pointing out is that keystones will only be included in solutions if they are check-tagged (so if you check Phase Acro but not Acro and click "Run", it will give you an error because the tree is disconnected).

If you think this is in a pull worthy state, I'll remove the things that don't do anything from the UI for the pull.

### "Optimize existing tree" use case

You can optimize an existing tree as follows:

1. Skill your tree in the general UI.
2. Open the skill tree generator window and make sure total points equals the number of skilled nodes (if you start it after skilling the tree and it is the first start, it should do that automatically, if not, change additional points until total points are correct or click "Reset").
4. Switch to "Advanced" -> "Settings" and select weapon class, off hand and tags fitting your build and main skill.
5. Switch to "Attributes", click "Load from tree" and then "Convert to pseudo attributes".
6. Remove attribute constraints that are unimportant for you.
7. Switch to "Pseudo Attributes" and remove constraints that are unimportant for you.
8. Make sure the automatically check-tagged nodes are only nodes you feel must be included in any case (like keystones and important notables).
9. Optionally check the "current tree as initial solution" check box. If similar trees are better, this will generally produce better results. If the tree needs to change heavily, it will generally produce worse results. If you keep it unchecked, make sure to reset the skill tree.
10. Click "Run" and wait.

Do note that trees are very specific, so if it is already an acceptable tree, it probably won't get any better outside of pathing between nodes for which "Skill Tagged Nodes" is suited much better.

### Pseudo attributes in the general planner

Since the support for pseudo attributes was already requested for the general tree planner (see #185) and my implementation would easily support it, I'd work on that before implementing the missing generator features if others find it useful.

My implementation loads the definitions of pseudo attributes with contained attributes and their conditions from xml files and evaluates if attributes are calculated into the pseudo attribute based on those conditions. See the [Data/PseudoAttributes](https://github.com/brather1ng/PoESkillTree/tree/treeGen/WPFSKillTree/Data/PseudoAttributes) folder for the xml files and the [TreeGenerator/Model/PseudoAttributes](https://github.com/brather1ng/PoESkillTree/tree/treeGen/WPFSKillTree/TreeGenerator/Model/PseudoAttributes) folder for the implementation.

### Screenshots

Attributes Tab
Pseudo Attributes Tab
Settings Tab

If you like to have a download link to the compiled program for trying it out, I'll try to get that up.


EmmittJ:
This looks really good! I did some testing on a couple builds, and it seems to work like a charm. I haven't found any bugs yet. Can everyone else pull this and mess around with it? @l0g0sys @SpaceOgre

MLanghof:
Currently, clicking "Reset" in the settings view deletes all the attributes in the list as expected, but it doesn't re-add them to the pool of mods you can choose to add. For example: http://i.imgur.com/3okNph3.png

brather1ng:
Fixed that and another bug related to it.

Thanks! Yes, it feels way better than what we had at first. I don't understand why datagrids are so unintuitive to use out of the box.

EmmittJ:
So been messing around with this and maybe I missed something, but why is the 'Automated' tab unable to be clicked lol

EmmittJ:
If you would like to remove the unused ui elements I will pull this a we can throw up a new release

brather1ng:
Unused/confusing stuff is removed from the UI now.

I am not sure whether having the 'Tagged Nodes' tab there is useful. It was useful for debugging but if you think it creates more confusion than it helps, I can remove that one as well.

EmmittJ:
Maybe remove/comment it out for now. It seems as though it should be too hard to actually implement that feature entirely into this UI (which is a good idea).

brather1ng:
Moving it entirely to the new UI sounds good. The code is implemented with the new stuff anyway, its just the old menu entry I kept.

I'll probably disable the useless settings when the 'Tagged Nodes' tab is selected and throw some explanation into that tab and then remove the old menu entry. But that is something for later, so I'll just comment it out for now.

EmmittJ:
I like that idea. Thank you for your hard work! This is seriously cool. Do you mind if I use your pull request description in the wiki section for the general public to learn how to use it?

brather1ng:
No problem, was fun to do this :)
Nope, feel free to use it.

EmmittJ:
Ah, something I just realized, so when running the test 'DijkstraUnconnected ' the test case fails do to no GraphNotConnectException.

brather1ng:
I did run those tests at some point ...
Probably changed something so it gets thrown at a different point. I'll look into it in a few minutes.

EmmittJ:
Sweet, thanks! I don't exactly what to mess with that stuff since I don't exactly have knowledge of the system as a whole.

PR #221, opened 2015-12-02
Fixes #220

### LINKED ISSUE #220: [Auto Tree Gen] Unarmed taking weapon nodes
if in settings you set your weapon as "unarmed" and tag with "attack" and say you want some pseudo fire damage, the program will put skills into things like "increased weapon elemental damage" and "increased fire damage with weapons", but "weapon" modifiers don't work for unarmed.

This was sent to me via reddit. I am just putting it up here.

brather1ng:
Well, that went faster than I expected.
- The not conditions were not read from the xml which lead to weapon nodes being taken if unarmed.
- "Twin Terror" was not taken because its attribute was missing from the xml.
- "Serpent Stance" was not taken because of a typo in the xml.

For the missing attributes, if you plan on answering him, "+x to maximum life" and "+x to maximum mana" are missing because they are explicitly blacklisted and "+x% to maximum y resistance" is missing because it's unique (check tag the node if you want the attribute).


EmmittJ:
Yep, I answered him.

PR #228, opened 2015-12-10
Added a simple Settings Menu

EmmittJ:
This menu currently controls node highlighting
This is mainly for people that have reported color blind issues. We could also turn this into a proper settings menu. 
This may not be the correct implementation of it, but I hope it is. 
@SpaceOgre @l0g0sys @brather1ng 
![Menu] (http://i.imgur.com/PSbZjIi.png)


SpaceOgre:
Looks good to me, but I'm not able to test it since my computer is in a closet (painting the computer room). Maybe just re-size the window to fit the content better.

And I think we should consider what else we could move here, but that can be done after this is merged.

l0g0sys:
With future changes in mind, I would not place Settings under File menu. File menu should contain only build related functions. In the future there should be also Properties... menu item which will allow to edit current build properties. Global application Settings should be under Tools or Edit.

EmmittJ:
Yeah, I wasn't sure where I was going to put it. The edit menu was the
second option. I will go ahead and update that.

Issue #226, opened 2015-12-09
Linux/others feature request


ccomly:
Would it be out of the question to compile this for linux/mac/others ?

couldn't get it running with Wine https://www.winehq.org/ , so either run it on windows through virtualbox or similar, or via team viewer to the host windows box (i play poe with steam in-home streaming to my linux laptop) 

i just figured since its open source, i dont know what kind of framework or libs ur using so i guess that could be an issue, or lack of demand.  Would be really nice to have it running natively on linux or atleast functional with wine.

Just a feature request :)

EmmittJ:
Well, this is written in C# using the .net library, so native operation
using our current technologies isn't possible.

ccomly:
all right thanks for the reply! ill keep trying with wine. I remember MONO was some kind of replacement for  .NET no? idk how compatible they are.

EmmittJ:
> The easiest way to describe what Mono currently supports is:
Everything in .NET 4.5 except WPF, WWF, and with limited WCF and limited ASP.NET 4.5 async stack.

http://www.mono-project.com/docs/about-mono/compatibility/

This is kind of a problem since this is exactly what we use in the skill tree. This would require an entire re-write of the current project. I surely don't want to do that

EmmittJ:
@brather1ng Is there any reason we updated to .net 4.5 other than async/await keywords? There is a library:  https://www.nuget.org/packages/Microsoft.Bcl.Async/ that adds these to .net 4.0

brather1ng:
I think async/await is the only thing we use from .net 4.5. If that package you linked also supports Visual Studio 2015, I would not mind going back to 4.0.

Else we could hack together a not really maintained 4.0 version by just removing the features that use async/await.

@ccomly Have you tried Wines .net 4.5 support? https://appdb.winehq.org/objectManager.php?sClass=version&iId=25478

ccomly:
i haven't tried that specific version with 4.5 will try it now.

EmmittJ:
If that wine package doesn't work then I would like to go back to .net 4.0
since I have had many people message about Linux and 'work' support lol

ccomly:
I tried a few versions, tried via regular wine / winetricks, crossover , playonlinux, after installing .net 4.5 i get the crashings

ccomly:
Not a developer/coder but i always fancy open source. Some times i can even fix stuff myself!

EmmittJ:
[...]

PR #232, opened 2015-12-11
Downgraded to .Net 4.0 for compatibilty with other operating systems

EmmittJ:
Updated Test Build to reflect Mana Per Level Changes
Updated packages to match
Fixed misc. errors that were caused by downgrade

This is a suggested fix for #226. I need everyone to build and run this on their systems to make sure there aren't any compatibility issues. 
@SpaceOgre @l0g0sys @brather1ng @MLanghof


brather1ng:
Everything seems to work fine for me (Win 10, VS 2015).
All async-using features and the settings stuff worked for what I tried.

EmmittJ:
I am on Win 10, VS 2012 just to add my specs

l0g0sys:
Work's on Win7, .Net 4.5.2 running with CLR v4.0.30319.
Though, for some reason menus are aligned with top-right corner instead of top-left. It looks really awkward.
menu-alignment-bug

EmmittJ:
Hmm, I am not having that issue.

EmmittJ:
Whoops, I am using VS 2013 Update 2, wasn't at comp to actually check what I was using.

SpaceOgre:
Sadly I will not have access to my computer for a few more days so won't be able to test. One thing though, I think it would be good to test run this on Linux before we commit to the downgrade, since the new packages might not work correctly.

EmmittJ:
I agree. I don't exactly have a Linux box set up, but I might be able to
get one running.

EmmittJ:
So, after running [this] (https://support.microsoft.com/en-us/kb/2468871) (KB2468871) I was able to run this successfully on Windows XP. I am not sure we can guarantee Linux support, but we can at least get people on XP back to using an updated skilltree.

EmmittJ:
Would you guys be cool if I pulled this?

SpaceOgre:
To be honest I don't feel like windows xp is a good enough reason for this
change. An OS that is no longer supported by its manufacturer shouldn't
really impact a new app.

EmmittJ:
True, but the thing is there are a lot of people that continue to pm about it. If WINE was updated this would be nice

EmmittJ:
I would agree though that it would make sense not to support it.

brather1ng:
If XP support is the only thing this would provide it doesn't make much sense to pull it imo.

Downgrading to .net 4.0 because many people can't upgrade their .net framework on the pc they want to use the app would be a valid reason. But I don't know how many people are affected by that.

EmmittJ:
That is another reason. I am not sure what our demographic is because
Github doesn't provide those stats, but I have a lot of people requesting
4.0 because they can't use it at work on 4.5 haha

EmmittJ:
I will close the pull request, but I will keep the branch up if someone thinks they can work on it.

Issue #219, opened 2015-11-21
Ascendancy support


MLanghof:
Well, [Ascendancy](http://www.pathofexile.com/ascendancy) is coming, and with it lots of new mods and a whole new... chunk of the skill tree.

My thoughts so far:
* The ascendancies' trees (I'll call them subtrees) will need to be toggleable, and with them node highlights, mouseover things etc.
* I'm curious how the website will do it, we should probably mirror the behavior.
* Enforcing the 6 point limit on those subtrees shouldn't be particularly challenging.
* Inclusion of this stuff in the skill tree optimizer and generator is basically just additional preprocessing.

EmmittJ:
Just a quick note, there is no need to limit the node count. We don't do
this currently.

brather1ng:
I would say the general UI part is not too hard. Just make it a toggable overlay were you see the Ascendany skill tree and can switch between the three for the selected class.

I see the problem at mirroring the behavior of the main skill tree for those smaller ones (assuming what the official website gives us is similar to what it gives for the main skill tree) because the current skill tree stuff is not written in a way were it could easily be changed to also produce the smaller ones (well, except copy-pasting the whole skill tree code and changing what needs to be changed, but I doubt that is a good idea).

But I guess it is too early to make assumptions on what the official website supplies us with. Maybe some refactoring of the skill tree code to make it more modular could be done at this point in time.

MLanghof:
> But I guess it is too early to make assumptions on what the official website supplies us with. Maybe some refactoring of the skill tree code to make it more modular could be done at this point in time.

I was thinking about that too. Compared to the website, we have the additional problematic of node highlights (what if a node behind the subtree is highlighted)? There are of course solutions for this (draw order?), but we'll have to cleanly separate the drawing/display code from the internal representation (those subtrees still have points allocated even if they're not being rendered).

brather1ng:
If the UI is properly done and those subtrees are entirely separated from the main tree that shouldn't be much of a problem. For example, you don't see highlights through the "Attributes" overlay. But, as you said, this will require separation of internal representation and the ui code, which is the reason the skill tree code as it is now won't work for subtrees.

brather1ng:
Well, things are getting a lot more interesting than I thought.

I guess a rather easy first step would be to ignore the Ascendancy classes themselves and just throw a button somewhere that enables taking unconnected nodes. For the ascendancy classes we will have to see what the officical JSON looks like.

[...]

PR #245, opened 2016-02-27
Support for additional class starts from Scion ascendancy class

brather1ng:
As discussed in #219. Should be easily adjustable for actual ascendancy class support.

Nodes now have an additional field for connections that should be visible. For Ascendant starts the start nodes of the class are added as invisible connections to the Scion root node.

Some remaining issues:
- After switching the additional class start or the class (to non Scion) there might be unconnected nodes. Refunding any node refundes all unconnected ones, so this isn't a big issue.
- No connections are shown to the additional class faces since the root node is not skilled. This could be solved by skilling the root node (if that isn't necessary anyway, see below). However, I don't know how compatible two root nodes are with the URL format (in case it is not necessary) and this would need to be subtracted anywhere the total node count is used.
- Depending on how the stuff is actually implementing in the official skill tree planner, the additional root node might need to be added.

So I guess it's best to wait with merging this until we have the official data. Wouldn't be of any use before that anyway.

### DIFF
--- a/WPFSKillTree/Model/Ascendancy/AscendantAdditionalStart.cs
+++ b/WPFSKillTree/Model/Ascendancy/AscendantAdditionalStart.cs
@@ -1,0 +1,16 @@
+namespace POESKillTree.Model.Ascendancy
+{
+    /// <summary>
+    /// Enumeration of possible additional starting points for the Scion through her Ascendancy class.
+    /// </summary>
+    public enum AscendantAdditionalStart
+    {
+        None,
+        Marauder,
+        Ranger,
+        Witch,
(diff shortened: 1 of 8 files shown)


EmmittJ:
Yeah, I will go ahead and wait to see what happens with the data. I like the idea of visible/invisible node links.

I do think we should skill the additional root node, but like you said, I am also not sure of how that will interact with the official URL. As a side note, this invisible link will give us a nice start for Intuitive Leap.

EmmittJ:
I am going to merge this, since the main tree structure hasn't changed

PR #256, opened 2016-02-29
Implemented user-defined attribute groups

Sectoidfodder:
Mostly implemented between MainWindow and GroupStringConverter.  

FormChooseGroupName is a simple text field.

PoEBuild only augmented with an extra property to allow saving/loading custom groups as part of a build.

### DIFF
--- a/WPFSKillTree/Utils/Converter/GroupStringConverter.cs
+++ b/WPFSKillTree/Utils/Converter/GroupStringConverter.cs
@@ -1,7 +1,9 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Text.RegularExpressions;
 using System.Windows.Data;
 using POESKillTree.Localization;
 using POESKillTree.ViewModels;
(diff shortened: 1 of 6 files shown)


EmmittJ:
Hmm, looks like I can't merge this, and I am currently on my phone just about to check some sleep. Sorry for any commits that broke this .

Sectoidfodder:
resolved conflicts - we both added properties to PoEBuild (CustomGroups and AscendantAdditionalStart), so wherever PoEBuild is created/copied there was a conflict of the lines we added to handle those.

Though now I'm getting a KeyNotFoundException in the merged code in SkillTree-Drawing.cs:

BitmapImage ascedancyNormalPImage = _assets[NodeBackgrounds["ascendancyNormal"]].PImage;

Any idea why?

EmmittJ:
It is possible you don't have the assets in your /bin/debug/data folder

Sectoidfodder:
Yup that was it.  Tested the customs groups w/ ascendancy features and everything seems to work fine.  Just need your stamp of approval.

Also realized I probably should've included some notes on how the groupings work:

> Everything is accessed via context menu from right-clicking the attributes area.

> Using a "#" in a group's name will result in it being replaced by the sum of all the numeric attributes in the group.

EmmittJ:
Alright, cool! I read through the code a bit, so I kind of got that. I didn't catch the part about the # in group titles though, that is sweet! 

I will go ahead and merge this.

PR #263, opened 2016-03-01
Removed remaining AscendantAdditionalStart related code; two fixes

brather1ng:
- Removed remaining AscendantAdditionalStart related code
- Ascendant "Path of the ..." nodes can only be directly skilled.
- Reenabled highlighting nodes by attributes. There is a context menu entry for highlighting and one for removing highlights as opposed to the previous toggle like behaviour.


EmmittJ:
Looks good! Thanks!

Sectoidfodder:
What did you need to remove the "lbAllAttr.PreviewMouseRightButtonDown += IgnoreRightClick;" handlers for?

I added those because user-defined attribute groups are handled via context menu activated by the right mouse button UP event, so any other functionality that happens on mouse button down would get in the way.

EmmittJ:
Wasn't that there to prevent the highlight function? There isn't a reason
to ignore the right down function if nothing is attached to it.

Sectoidfodder:
Right, I see he moved the old highlight stuff into the context menu.  Forgot I had the old highlight stuff disabled when I started working on this.

brather1ng:
Overwriting those handlers prevented the right clicked item from being selected. The selection is one way to know which item was right clicked, which is necessary for highlighting the corresponding nodes.

PR #264, opened 2016-03-01
Tree comparison support for attribute groups

Sectoidfodder:
Sums attribute deltas in the group title while "compare trees" is on.

Added necessary Refresh() call to one place in MainWindow.


EmmittJ:
Thanks for the update, looks good!

Issue #251, opened 2016-02-29
"Skill tagged nodes" crash


Fasbi:
http://pastebin.com/Uw8NxQ0i

Even when selecting only one node in front of the starting point of the class - it crashes when I try to use "Skill tagged nodes".

EmmittJ:
Error in MinimalSpanningTree.cs - Line: 68 - `nodes.UnionWith(path);`

MLanghof:
Will look into it.

brather1ng:
This was fixed and can be closed.

PR #266, opened 2016-03-01
Fix for #251

MLanghof:
Fixing #251.

The algorithm now ignores all Ascendancy nodes as potential pathing nodes.
This also makes sense, because players can *really* just skill the Ascendancy nodes themselves.

It would make sense to disable the tagging of Ascendancy nodes altogether, but I don't want to make that decision myself.

### DIFF
--- a/WPFSKillTree/TreeGenerator/Algorithm/DistanceLookup.cs
+++ b/WPFSKillTree/TreeGenerator/Algorithm/DistanceLookup.cs
@@ -178,11 +178,19 @@ namespace POESKillTree.TreeGenerator.Algorithm
             _distancesFast = new int[_cacheSize, _cacheSize];
             _pathsFast = new ushort[_cacheSize, _cacheSize][];
 
-            FullyCached = true;
-            foreach (var node in nodes)
+
+            try
             {
-                Dijkstra(node);
+                foreach (var node in nodes)
+                {
+                    Dijkstra(node);
+                }
             }
+            catch (GraphNotConnectedException)
+            {
+                throw new InvalidOperationException("The graph is disconnected.");
+            }
+            FullyCached = true;
 
             // No longer needed.
             _distances = null;
@@ -289,6 +297,13 @@ namespace POESKillTree.TreeGenerator.Algorithm
             // Target node was not found because start and target are not connected.
             if (target != null)
                 throw new GraphNotConnectedException();
+
+            // Check if all nodes were reached
+            foreach (var otherNode in _nodes)
+            {
+                if (_pathsFast[start.DistancesIndex, otherNode.DistancesIndex] == null)
+                    throw new GraphNotConnectedException();
+            }
         }
 
         /// <summary>
--- a/WPFSKillTree/TreeGenerator/Solver/AbstractSolver.cs
+++ b/WPFSKillTree/TreeGenerator/Solver/AbstractSolver.cs
@@ -369,7 +369,9 @@ namespace POESKillTree.TreeGenerator.Solver
                             // Only add nodes in the subsettree if one is given.
                             || Settings.SubsetTree.Count > 0 && !Settings.SubsetTree.Contains(node.Id)
                             // Mastery nodes are obviously not useful.
-                            || node.IsMastery)
+                            || node.IsMastery
+                            // Ignore ascendancies for now
+                            || node.ascendancyName != null)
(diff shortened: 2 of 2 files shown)


brather1ng:
It always shows the unconnected-Popup for what I tried (with and without your second commit).

I think that safeguard from your second commit won't work with the AdvancedSolver. Relevant nodes can very well be unconnected, e.g. you have life as a target attribute but not check-tagged Blood Magic. Those are currently filtered out based on the information from DistanceLookup.

> It would make sense to disable the tagging of Ascendancy nodes altogether, but I don't want to make that decision myself.

Imo all highlighting and tagging of Ascendancy nodes can safely be disabled.

MLanghof:
> It always shows the unconnected-Popup for what I tried (with and without your second commit).

It crashed before the second commit for me if I tagged nodes inside an ascendancy tree.

> Relevant nodes can very well be unconnected, e.g. you have life as a target attribute but not check-tagged Blood Magic.

Hm, that's correct, I didn't think of that.

> FullyCached = true; being before the Dijkstra-calling-loop is relevant

Argh, my bad yeah. Makes perfect sense. I didn't catch that (rather glaring error) in my short tests.

> (e.g. it doesn't use the additional starts from Scion)

Shouldn't that be pretty easy to "fix" by requiring people to spec into the respective Ascendancy? If we directly connect the classes' (actually skillable) two starting nodes with their respective Scion "exit" conditionally, that should do the trick for the solver, no?

brather1ng:
> It crashed before the second commit for me if I tagged nodes inside an ascendancy tree.

Hm, couldn't reproduce that. But that should be no problem if we disable tagging anyway.

> Shouldn't that be pretty easy to "fix" by requiring people to spec into the respective Ascendancy? If we directly connect the classes' (actually skillable) two starting nodes with their respective Scion "exit" conditionally, that should do the trick for the solver, no?

I think those nodes are actually connected. I don't know why it doesn't work even if the connection ascendancy nodes are skilled. But I didn't spend any time looking into it, so it might be trivial to fix,

MLanghof:
Well it certainly should be an easy fix. I can just undo the "error checking" and prevent the ascendancy tagging, which should deal with the issue.

EmmittJ:
I think removing ascendancy tagging is a good idea. If someone really
wanted the scion ascendancy stuff and start at a new place they can just
skill 'Path of the ...'. The algorithm should work since the starting nodes
are a possible 'out'

MLanghof:
Yeah. However, I'm having a hard time with preventing node tagging for ascendancy nodes because the node highlighter is interfaced in many places in many ways and (sort of) treats both tagging and highlighting pretty much the same way. Plastering checks all over the place seems error-prone especially for the future.

I fear I'm not thinking straight after working on my bachelor thesis for 12 days in a row, and it looks like my commits ended up being rather broken as a result. Since brather1ng seems to be engaged, I think I'll just leave the stage for him if that's fine. I don't think I should be doing major code reworks at the moment.

EmmittJ:
Dang. You should just be able to prevent the right click from working on
ascendancy nodes. That is, although, a quick fix.

MLanghof:
Well yes that's trivial, but there are more ways to tag nodes (namely, the menu entries for converting highlights to tags).

brather1ng:
I would've thought to disable highlighting of ascendancy nodes entirely. There isn't really any use of having those highlighted, you know where they are anyway.

EmmittJ:
Ah, shouldn't they all use the same method? Right clicking and converting
highlights should all call the same check function?

brather1ng:
There is the NodeHighlighter class in which 1-3 methods have to be adjusted. That should be all that is necessary.

Do we only want to disable check-/cross-tagging or all highlighting for ascendancy nodes?

@MLanghof I do have some time now. I'll see how far I can get so you can give your brain a break ;)

EmmittJ:
Probably just checking, because it would be nice to be able to
highlight/search for nodes in ascendancy trees

MLanghof:
> Right clicking and converting highlights should all call the same check function?

Well, unfortunately they do not, the check would have to go in at least three different functions no matter at what level it would be inserted (UI/Controller/Highlighter class).
Thanks brather1ng!

EmmittJ:
Ah, alright. Thanks for letting me know. I haven't really looked into the
system you guys have created.

PR #268, opened 2016-03-01
Solver fixes

brather1ng:
In addition to what was done by @MLanghof in #266:

- Disabled check- and cross-tagging of ascendancy nodes.
- Fixed Skill Tree Generator crashing on window opening.
- Renamed crit multi pseudo attributes to the new naming scheme.

Bugs I know of but don't have time to fix atm (probably won't get to them over the next few days):
- Ascendant additional start nodes are not used for the solver (even if the "Path of the ..." node is skilled). This is because the connections between the start nodes and the Ascendant nodes are not bidirectional so the start nodes are thrown out of the search space because they don't have the "Path of the ..." node as a neighbor.
- Ascendancy nodes count to the node limit in the Solvers.

### DIFF
--- a/WPFSKillTree/Data/PseudoAttributes/Crit.xml
+++ b/WPFSKillTree/Data/PseudoAttributes/Crit.xml
@@ -34,9 +34,9 @@
 			<Conditions><WeaponClass>Staff</WeaponClass></Conditions>
 		</Attribute>
 	</PseudoAttribute>
-	<PseudoAttribute Name="#% increased Critical Strike Multiplier" Group="Critical Strike">
-		<Attribute Name="#% increased Critical Strike Multiplier" />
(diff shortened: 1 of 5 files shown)


EmmittJ:
So pull this, close the other? 

As for the start nodes not being bi-directional, I didn't know of a better way to prevent the distance algorithm to ignore the ascendancy tree.

brather1ng:
MLanghof's commits from the other are included here. So, yes, I guess.

Yea, it's ok. The solver was just not setup for directed edges. It needs more time than I have atm but it shouldn't be too bad. And it's not a big problem anyway, one only needs to skill the actual start node and it will work.

EmmittJ:
Alright, cool. Thanks for helping with this.

EmmittJ:
Oh, that is weird haha it merged both requests. That is pretty cool.

PR #269, opened 2016-03-02
Minor UI update

Sectoidfodder:
Replaced default alert box with Util.Popup


EmmittJ:
Is the popup info window a metro window?

EmmittJ:
Also, when displaying messages or any text you should use the translation
system we have.

PR #273, opened 2016-03-03
Added implicit attributes, fixed base mana

Sectoidfodder:
Added "\<expression\> with all charges" implicit attributes derived from "\<expression\> per charge" and "+X maximum charge" attributes.  They're useful for summing up the relevant stats in custom groups.

I'm undecided whether the original "per charge" attributes should be hidden/removed in order to reduce clutter.

Base mana at level 0 fixed.  It probably wasn't updated to account for the mana buff a few patches back (always 40 mana at lvl 1, but 6 per lvl instead of 4).

### DIFF
--- a/WPFSKillTree/SkillTreeFiles/SkillTree.cs
+++ b/WPFSKillTree/SkillTreeFiles/SkillTree.cs
@@ -51,7 +51,7 @@ namespace POESKillTree.SkillTreeFiles
 
         public static readonly Dictionary<string, float> BaseAttributes = new Dictionary<string, float>
         {
-            {"+# to maximum Mana", 36},
+            {"+# to maximum Mana", 34},
             {"+# to maximum Life", 38},
             {"Evasion Rating: #", 53},
             {"+# to Maximum Endurance Charges", 3},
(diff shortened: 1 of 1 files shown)


EmmittJ:
With the base mana changes the the tests may fail, which means the test build needs to be updated.

Sectoidfodder:
TestBuild mana was off by 2 as expected from my edit, but two other parameters - life and accuracy - were also wrong.  I'm pretty sure I haven't done anything to those.

I don't have any experience with the item import and build stat calculation process, so not sure what I can do other than forcing builds.xml to match the "wrong" output results instead.

EmmittJ:
It isn't the correct solution, but that is because the tree needs to be remade and reevaluated as per my note on the commit. We can't have a test build on the current data without actually being able to have a build :P

l0g0sys:
Yeah, just make it run through tests so release can be made if needed.
You can pretty much delete all old testing builds. Lot of stuff changed 
so they have to be deleted anyway.
I'll be adding new test builds once new league will start and i start 
playing with fresh char.

EmmittJ:
Sweet, thanks.

PR #278, opened 2016-03-04
Fix #274, attr groups message box cleanup

### LINKED ISSUE #274: Skill tree comparision: difference per attribute not shown
The differences in attributes between two skill trees are not shown if compared via "Skill tree comparision" (works in the latest release).

Sectoidfodder:


EmmittJ:
Looks good, thanks!

Issue #234, opened 2015-12-12
[request]Screenshot Save as ...


Hixxie:
First, I don't know where the directory is that images are saved to. I've looked and can't find it under my documents, the program folder, application data, etc.

Second, users should be able to save file as (dialog window) to change the file type (jpg, png, bmp, etc) and the directory.

brather1ng:
It is saved to your clipboard, you can open any image editing program (even Paint is enough) and press Ctrl+V and it should be pasted there.

But yes, removing the need to open a different program for saving screenshots would definetely help.

Pecral:
I've added this now for jpg and png.

The problem is, if the user didn't select any node, there will be no file dialog at the moment.
Although this is intended, it could be a bit confusing if nothing happens after clicking a button - there should be a little information box at least.

You can ignore the first commit.. That's my first commit ever and I didn't know that a commit on a forked repository is automatically mentioned here.

SpaceOgre:
@Pecral How do you feel about this? Is it in a state where others might use it? If so please do a pull-request.

Pecral:
Sorry for the late response!
I've created a pull-request now.

EmmittJ:
This has been added with #294

PR #294, opened 2016-03-22
Added feature #234

Pecral:
A save file dialog will open if the user wants to take a screenshot of his build. There should be at least one node skilled in the build, otherwise a message box will show that the user has to skill something.

The default name of the file is the current build's name concatenated with the number of used points e.g. "CI FP - 95 points used.jpg"

Edit: Mh, the "used" in the filename is a little bit pointless I think. I'll remove that.


EmmittJ:
Looks good, and it works flawlessly. Thanks!

PR #296, opened 2016-03-25
German translation

Pecral:
1. Changes to localization references in XAML files:

 - MainWindow: Moved all localization templates which were in the "Header" attribute of a MenuItem to a separate "MenuItem.Header" element. Otherwise the gettext-plugin won't create an entry in the localization .pot-file. (Well, this was the case for me, I hope this wasn't just a configuration issue)

 -  MetroMessageBoxWindow: Moved the declaration of the button's content-text (Yes, No, Ok, Cancel) to a nested l:Catalog so that an entry in the .pot-file can be created.

 - DownloadStashWindow and CraftWindow: Moved the window title to a l:Catalog in an "MetroWindow.Title" element so that an entry in the .pot-file can be created.

2. The layout of the settings menu is now a grid with two columns. The width of the first column is set to "auto" to prevent issues with long localized labels.

3. Added german translation

### DIFF
--- a/WPFSKillTree/Locale/de-DE/Help.md
+++ b/WPFSKillTree/Locale/de-DE/Help.md
@@ -1,0 +1,52 @@
+﻿# Grundlagen #
+
+Willkommen zur Hilfeseite von PoESkillTree! Wir werden damit starten, die Hauptfunktionen des Programms zu erklären.
+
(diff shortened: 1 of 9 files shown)


EmmittJ:
Thanks for the translation!

PR #304, opened 2016-04-11
Added additional app specific strings to German.isl

brather1ng:
Together with what @SpaceOgre already commited, this seems to be all that is necessary to support a  German installer. Though, I don't know much about our installation stuff, so I haven't tested it.


EmmittJ:
Could you add `Name: "de"; MessagesFile: "compiler:Languages\German.isl"` to `WPFSKillTree/release.iss`.

This should add it as an option for installation

brather1ng:
That was done by @SpaceOgre in 84fe39abb1f539f952661529290b197016403f9e (well he used "de_DE", which is what the translation folder is called)

EmmittJ:
Ah, sorry. I missed that commit. Thanks!

SpaceOgre:
Ah, nice. Was not sure if this part was needed since I have never used the installer... :)

EmmittJ:
Yep, this is the install instructions in the native language

PR #307, opened 2016-04-17
Improvements to the Tree Generator

brather1ng:
These are the results of very sporadic work over the last half year. So I might have forgotten something.

The improvements mainly result in two things:
- Performance improvements: The SteinerSolver (the one behind "Skill Tagged Nodes") is now nearly instantaneous (<300ms in Debug mode) and the AdvancedSolver is a bit faster (around 10-20%).
- Better results: The SteinerSolver is now guaranteed to return optimal results in many cases because the search spaces gets really small for realistic inputs (dimensions of less than 10 in many cases). This leads to complete search space exploration by the genetic algorithm. For the AdvancedSolver the reductions that lead to this are not enabled atm. But they would reduce the search space dimension by around 100 (from ~300 to ~200) if enabled.

These improvements are mainly caused by reduction tests I implemented. There is now a class that preprocesses problem instances using different reduction tests (I moved the reductions already in place in the current version there, too). See the class `TreeGenerator.Algorithm.SteinerPreprocessor.cs` and the `TreeGenerator.Algorithm.SteinerReductions` directory for more information.

The SteinerSolver is now at a point where I don't thing investing more time into it is worth it: the calculation feels instant and the results should be optimal in all cases (if someone finds cases where the result is not optimal, that might very well be a bug in my code).

I initially implemented the reductions in the hope that they improve the results of the AdvancedSolver, but from my experimentation, they do not (it's totally possible that they do, however, the AdvancedSolver is too random to notice any significant difference atm). I will experiment more with it in the future.

Some not directly related notes:
- I tried a bunch of different things, so most of the (not UI-related) Tree Generator code got refactored, resulting in better performance or (subjectively) better design.
- Someone in the Reddit thread for the Tree-Generator-release suggested that we can very easily do this with an SMT solver. Trying that was what first started my work on this branch. But that proved to be far too slow even for the SteinerSolver so I didn't even try it for the AdvancedSolver. That might have been caused by the logical encoding I choose, which basically was the first somewhat easy one I found. It might be possible to get it to an acceptable performance but I didn't try it out any further because encoding things like paths and connectivity in graphs as logical formulas is anything but easy.
- The tests I implemented are mostly taken from scientific papers about the Steiner tree problem. From what I've seen, our problem instances for the SteinerSolver are on the easier side of things, so guaranteeing optimal results in reasonable time is totally possible with proper algorithms (that are far more complex than our genetic algorithm) and the more complex reduction tests that I skipped. These reductions alone would probably already solve most of our instances (i.e. reduce the search space to 0 so there is only an optimal solution left).


EmmittJ:
Thanks for the update! Anything that you went us to mess with before merging?

brather1ng:
Well, someone else than me playing around with "Skill tagged node" a bit certainly wouldn't hurt. But I think I tested it enough that it can be merged now and possible bugs can be be fixed as they occur.

EmmittJ:
I have to say it is a hell of a lot faster. This is some amazing work, thank you.

MLanghof:
I see. I was wondering if there would be some interesting progress towards the Z3 implementation (in particular I was wondering how the mapping to the logic would work), but I can imagine that it's introducing a ton of additional (redundant) dimensions to the problem that the algorithms in the "original" implementation (MST, Dijkstra) are covering intrinsically.

It's also nice to know that our initial "handwavey" search space reductions are now replaced by scientifically funded ones. Nice work!

brather1ng:
[This](https://github.com/EmmittJ/PoESkillTree/blob/155598f3f85d896d72694a0dea8b438f865ad02c/WPFSKillTree/TreeGenerator/Solver/SteinerSmtSolver.cs) is the latest version of the SteinerSmtSolver (using Z3) before I removed it.

Most of what it does is making sure that the encoded graph is connected ("each target node has to have a subgraph which only consists of a path to the next target node" was the encoding if I remember correctly). The sum of the edge weights of all edges taken in at least on subgraph is the function to be minimized with that encoding.

Overall, this easily leads to a few thousand constraint statements (O(|target nodes| * |edges|), I think) and solving it takes far longer than our approximation. Extending this encoding for the AdvancedSolver probably makes it far more complex so I dumped the idea entirely.

PR #322, opened 2016-05-15
Making the Equipment tab usable

### LINKED ISSUE #155: Quality for Item Crafting
We should be able to add quality (whetstones and armour scraps) to items while crafting.

### LINKED ISSUE #199: Save Equipment
Atm we only save items added to a character from "Download & Import Items", any item changes done from the Equipment tab is not save to the character.

As I see it we have two options:
1. Save the equipment instead of item json string, as a list of serialized items.
2. Convert the equipment to json string when saving

### LINKED ISSUE #203: Stats do not update
After crafting a new item and placing it into my character equipment it does not update my defensive stats under character sheet.

### LINKED ISSUE #302: Passive Tree/Equipment Tabs
I'm not a fan about of how the Passive Tree/Equipment Tabs looks/are placed. Does anyone have a suggestion on how they could be changed/moved?

On the one hand I get the sentiment behind it since they toggle what we see in the "Editor" area but I'm wondering how many are using the equipment tab and if it might not be better to move them to the menu instead. A part of this is also because the equipment stuff is not working 100% either.

### LINKED ISSUE #324: reset button doesn't reset bandits
They will stay assigned to whatever you had before, i.e. from loading a saved skill tree.

brather1ng:
Our Item editor/Equipment tab is in a nearly usable state so I decided to fix and update it.

**Edit: A changelog can be found [here](https://github.com/EmmittJ/PoESkillTree/pull/322#issuecomment-224922194)**

This is still work in progress but I wanted to give you the possibility to discuss the stuff I am planning to do.

Todo list:
- [x] Rewrite DownloadItemsWindow: Changed to a design that doesn't look as quick & dirty as before. Item data can be pasted into a field instead of loading it from a file like before.
- [x] Up to date affix and item base lists:
  - The new parsers are located in a new subproject.
  - The information in the lists is in a format that is identical to in-game. Previously mods had to be aliased and were generally not identical to their in-game counterparts.
  - Affixes are extracted from exilemods, base items from the Wiki.
  - [x] Unknown bases (mainly Talismans and Demigods) are handled on a best effort basis to guess their ItemType.
  - [x] The parsers should update the project assets like the UpdateDB project does it.
- [x] Rework image loading:
  - [x] Extract images outside of main program (what the old ItemAssetDownloader does, that class can partly be used)
  - [x] Rework base item image loading so it is stored in the BaseItem instance and missing images are loaded on-the-fly from the wiki (only on item creation)
  - [x] Load images for imported items on-the-fly through the url that is included in the item data and save them in the asset folder once loaded.
  - [x] Merge UpdateEquipment into UpdateDB and refactor UpdateDB to the same extensible architecture.
  - [x] Load images as part of the packaging for release.
- [x] Rewrite "Download & Import Stash" the same way as the item data version.
- [x] Use MahApps.Metro style dialogs
  - [x] Provide a custom style for a consistent close button and use that for the already rewritten "Download & Import" dialogs
  - [x] Replace Popup/MetroMessageBoxWindow with those
  - [x] Update all existing dialogs to use the new style.
  - [x] Replace progress windows by MahApps' ProgressDialog (#329)
  - [x] Wiki article or something on using the new stuff.
- Smaller features:
  - [x] Item base selection in CraftingWindow consists of three steps instead of two: ItemGroup -> ItemType -> ItemBase.
  - [x] Item tooltip of crafted items shows level and attribute requirements.
  - [x] Equipped gear can be saved to stash and the item drag & drop has been greatly improved.
  - [x] Support for item quality (fixes #155)
  - [x] Better look for the Passive Tree/Equipment tabs (fixes #302)
  - [x] Better naming for the *Armour, *ArmourEvasion, ... item types in crafting window's type list
  - [x] Master signature mods (like "Hits can't be evaded")
  - [x] Save imported stash tab list (per league)
  - [x] Adjustable reset button (fixes #324)
  - [x] Separate gem file that contains properties not extracted from the wiki (supported weapon types, required hand, ...)
- Major bugs:
  - [x] Equipment is not saved to item data (fixes #199)
  - [x] Equipment does not apply to defense/offense (fixes #203) (was caused by incorrect naming of mods and is fixed except for some issues with exilemods)

Known bugs:
- (won't fix for now) Master crafted flat defense is not applied if it's not the natural defense of the base item.
- (won't fix for now) The accuracy suffix is listed twice in the crafting window: normal tiers (local) and master crafted (global).
- (won't fix for now) loading an inventory into the stash stacks items on top of each other
- (won't fix for now) Two-Stone Ring icons are all the same as different icons for the same item type name are not supported

The stuff I already did also includes cleaning up the item related code for less redundancy and better understandibility. I also fixed a bunch of small bugs that weren't yet reported.

I might do other related stuff later (edit/copy item, buff support, updates to Compute.cs, Uniques, Enchantments and Corruptions and support for Gems and Sockets, roughly in that order) but I feel that all of the stuff mentioned above is important to bring the item stuff to an acceptable level.

The diff is blown up because the affix and item lists are included and because I removed the NuGet packages from source control.


EmmittJ:
> Images for uniques are missing. Maybe load them on-the-fly through the url that is included in the item data and save the images in the asset folder once loaded.

We should be careful with "on-the-fly" here. This is the "Offline Planner", so we need to be weary of those that do actually use it without an internet connection. It would seem weird if they were loading item data without an internet connection, but it is possible. 

The other points make a lot of sense, and I am thankful that you are willing to fix up so much code. Is there anything specific you would like help with?

brather1ng:
> We should be careful with "on-the-fly" here. This is the "Offline Planner", so we need to be weary of those that do actually use it without an internet connection. It would seem weird if they were loading item data without an internet connection, but it is possible.

Yes, it would need to handle connection errors by showing the base item's image or a default image. An option in the settings to disable this kind of loading would be useful, too.

> Is there anything specific you would like help with?

I am half way done with the download stash window and then there's the image loading stuff I'll do after that. Everything else should be small enough that I don't think it's worth dividing it up.

One thing that just crossed my mind, is it possible to download the item images as part of the release script? It's probably not a good idea to include them in the repository. I would implement the image loading as part of the affix and item list project. The script would simply need to call the exe of that project with some argument (updating the affix and item list on release wouldn't be a bad idea either).

> edit: sorry this took so long to get back to you I haven't been at a computer lately. 

No problem.

EmmittJ:
Yes, it would be possible to load the images as part of release, but I
think we stayed clear of that because it would create a massive download
for a lot of people. The UpdateDB project gets called which loads the
newest skill tree data, so it shouldn't be too hard to have it update the
affix/item list

brather1ng:
I only thought about the images for the base items. That would be under 10 MByte, is that already too much?

EmmittJ:
As long as people are alright with the double download size, I don't think
it is a big deal. It does mean that for people that want to use the item
editor that they have a pretty big initial download.

brather1ng:
I wouldn't mind if downloading the images would be an installer option or an option on first startup. But I don't think having it only as a menu entry, as it is now, is a good idea. I for one didn't know we already support item images until I randomly clicked on that entry.

EmmittJ:
If people complain then there is always the next release to take it out. I
would say go for it. I can also look into it once you are done. You
basically can add a flag to UpdateDB and add it to the release.xml file.
Line 275 there will show you how it works.

brather1ng:
Well, that's easier than I thought. It's probably best to merge my UpdateEquipment project into UpdateDB anyway.

EmmittJ:
Yep! Super simple

brather1ng:
What do you think about replacing our current dialog windows with MahApps style dialogs?
mahappsdialog

This also has the upside that there is already a system for calling dialogs from non-GUI code, so we won't have these `Popup.Error()`-calls everywhere in logic code.

EmmittJ:
They look awesome, but we will need to make sure they all have a way go be
closed. Seems obvious, but the screenshot doesn't really give a good feel
off that. Does clicking the greyed out background close the dialog?

brather1ng:
They don't seem to have a way to be closed built in. In this case only the OK button closes it.

We should probably use a consistent style with such a button in the bottom right corner and a bit more highlighting than there is atm.

EmmittJ:
Yep! I would completely agree. Something along the lines of 'close' or
'cancel' would do the trick.

brather1ng:
The built in message window looks like this:
errordialog
That button makes it a lot more obvious.

EmmittJ:
That is very nice and a big improvement with just some simple highlighting.

brather1ng:
Since I just added another NuGet package that blows up diffs (we can use proper logging now), is it ok if I put them into .gitignore? They are automatically restored on first build so it's unnecessary do have them in source control.

EmmittJ:
Yeah, they are added to the dependencies, so there isn't a reason to add
then to source control.
On May 22, 2016 3:27 PM, "brather1ng" <notifications@github.com> wrote:

Since I just added another NuGet package that blows up diffs (we can use
proper logging now), is it ok if I put them into .gitignore? They are
automatically restored on first build so it's unnecessary do have them in
source control.

—
You are receiving this because you commented.

Reply to this email directly or view it on GitHub
<https://github.com/EmmittJ/PoESkillTree/pull/322#issuecomment-220860421>

brather1ng:
> −443,771 lines

There we go ^^

EmmittJ:
Nice haha

EmmittJ:
https://github.com/EmmittJ/PoESkillTree/pull/322/commits/b22e92df00d6037d4dcb7c76ffac7b1b58e14357

Missing images really shouldn't be downloaded on-the-fly. The user should need to request the program to download said images. I am not trying to be rude, but we are the 'offline' planner ;)

brather1ng:
Wherever you comment went, it can be changed in the settings s only downloading images on-the-fly on request is a matter of changing the default value of that setting.

EmmittJ:
Yeah, sorry. The commit log didn't log completely, and I made that comment.
Good work so far, thank you!

brather1ng:
The item images are now loaded as part of the release script.

I merged UpdateEquipment into UpdateDB and changed how the gem loading works: it uses the gem lists from the wiki to determine which gems to fetch rather than the update gem already in list/add a single gem/merge gem lists program arguments we had before. That should cover all use cases and it is no longer necessary to add each new gem manually. I don't know how the list was updated before so if I missed something, tell me :)

The dist-update.bat is gone as I figured it is more useful if the version controlled lists that are updated through UpdateDB now (gems, base items, affixes) are updated individually (or all at once if you want) by specifying command line arguments (see UpdateDB.Program.cs for the documentation on that). If a dist-update.bat one actually still has a use case, I can add it back.

Two things I had to change in the release.xml to get it to run on my machine:

1. The path to the dll that contains the CodeTaskFactory things is different on VS 2015 so it should now select the one that exists. It should still work on older Visual Studio versions but I can't test that.
2. I had to change the platform from `x86` to `AnyCPU` because otherwise it could not compile the submodule projects (RangeTree, WpfRangeControls). From what I read about that platform, it should be better anyway since it adjusts between 32 and 64 bit mode depending on what the computer it runs on supports.

EmmittJ:
1) It should be fine. I will test this later
2) I always updated the submodules to include the missing platform. 

Quick note on gems: they were manually added because of the properties they had like MainHand, OffHand, DualWield, etc.

brather1ng:
> Quick note on gems: they were manually added because of the properties they had like MainHand, OffHand, DualWield, etc. I recently pulled all the gems and manually updated them (few releases ago). 

Oh, I see. I think it is best if those properties are loaded from a different (manually updated) file. That should be much easier to maintain as it would only contain information about the gems that actually have those properties (only attack skill gems).

l0g0sys:
Gems were added manually because, it's possible to have wiki page for a 
gem and not having gem in list. Those are kept updated separately as 
wiki pages. There could be stale links in list to non-existing gem pages 
for example. And you still have to check entries in XML file after 
adding anyway to set correct flags for gems which are not parsed from 
gem page. Not to mention that new gems support should be properly tested 
before going into release.

Please let Items.xml as it is now. Don't add another "meta" file for 
those special attributes of gem usage. There is no real need to have 2 
files to keep updated. If new gem is added to Items.xml, it won't be 
added automatically to second "meta" file anyway. So having second file 
doesn't solve anything at all.

dist-update.bat was just batch file to call UpdateDB.exe to update 
Items.xml and maybe other future tools for update of application 
databases. So you just had to click it to get all update stuff done and 
didn't have to open CLI and execute binary manually. It was there to 
simplify the task of updating stuff nothing else.

Not sure if there was need to mess with CPU architecture. Never had 
issue compiling solution as each project has own CPU platform settings. 
AnyCPU architecture may have some interoperability issues with Win32 
legacy unmanaged DLLs (that's what i was reading about it). Idea of 
AnyCPU platform isn't much supported even by Microsoft itself (check 
newest mobile device development how-tos, they never suggest to use 
AnyCPU, but rather x86/x64/arm/... separately. It's way better to make 
two separate builds for x86 and x64 to have fully benefit from 
optimization for these architectures. Even .NET itself is released in 
x86/x64 separate packages and not using AnyCPU. I didn't find any 
benefits in AnyCPU yet just some interoperability limitations.

Anyway, I'll get rid of those git submodules asap, when i get time slot 
for it. It ended up exactly as i predicted it. I'd rather have them 
included in project directly than having everyone to deal with git 
submodules mechanics. We already saw "Cannot build solution, 
sub-projects missing", when someone wanted just evaluate build by simply 
downloading source code directly without Git tools and doing all that 
submodule voodoo stuff to get it work.

brather1ng:
I think the gem list is much easier to maintain if the information that can't be extracted from the wiki is kept in a separate (manually updated) file and everything else is in a file that is only automatically edited. You know exactly what can't be extracted from the wiki: required weapon and the like (that only affects non-support attack gems and two other gems) and legacy gems no longer in the wiki lists (which is only Item Quantity atm). As a matter of fact, I found a bunch of attack gems that were missing their manually set information.

> Not to mention that new gems support should be properly tested before going into release.

Well, nearly every unsupported gem listed at the bottom of ItemDB.cs is in the gem list and most (all?) new ones aren't in that list but partially won't work, so this wasn't done before either.

If you insist, I can go back to the old gem list model, but I will not maintain that list.

> dist-update.bat [...]

I can add that back and add item list and affix list to it, I had that done (but decided to remove the file altogether) anyway

> Not sure if there was need to mess with CPU architecture.

For me, it was necessary, dist-release doesn't run without. I don't really care if we use x86 or AnyCPU, it's just that it doesn't run with x86.

It's probably best to compile the submodules to dlls and throw those into the repository instead of having them as submodules. That should make it possible for me to run dist-release with x86.

l0g0sys:
Yeah, please roll it back as it was. It worked without issues. You add 
gem however you wan't (manually/updatedb.exe/write you own tool if you 
want), add special flags to it's entry and that's it. I've made it to 
work this simple way. No need to mainain any other file.

You can actually override gem entries in own local XML without need to 
mess with released XML file. This allows you to work with whatever data 
you want and you won't loose them ever, once you update to new release. 
There is no data loss in using UpdateDB tool either.

I think you misunderstand meaning of that XML file. It's not file of 
extracted data from Wiki. It's database file for our application. 
UpdateDB is extra tool i've made, to make some tasks easier, which i can 
delete from project at any time and nothing will be affected due to XML 
file still being there with it's defined structure and API. And anyone 
can write whatever tool to populate it. In no way it revolves around 
UpdateDB tool.

So, please, don't downgrade it to some lame Wiki extracted data file :) 
Remember that, one day Wiki may be gone. Don't rely on it too much.

I don't understand what you mean by old gem list model.

l0g0sys:
I'd really like to have Items.xml and it's API (ItemDB) independent of 
Wiki and not getting associated with it any way. With time I'd like to 
move lot of other information there, which never will be extracted from 
Wiki (like constants for life per level etc.). Maybe i didn't choose 
good name for class/file, but I think now you can understand what's 
purpose of it.

brather1ng:
Then you probably don't like that I renamed Items.xml to GemList.xml and moved it to Data/Equipment either (as it is now, it has absolutely nothing to do with items). If stuff is added to ItemDB in the future, it would be best to not keep it all in a single huge file, imo. ItemDB already supports merging so that shouldn't be a problem.

I'll remove the GemProperties.xml and change UpdateDB to not overwrite but update the file again (it would be identical to before then). Should that only update the gems that already exist or can it load all gems?

I changed UpdateDB to support arbitrary data loaders (gems, affixes, base items, ...), so I would have to reimplement all ItemDB/gem related options. Which of those options are necessary? Updating/Adding a single gem probably is useful, even if it uses the wiki gem lists. What is with merging an existing file (`/M`) and with changing the file name that is updated (`/F`)?

EmmittJ:
I understand the the old list wasn't compiled from the Wiki, but they have
worked hard getting the information up-to-date and have a python bot that
updates the information directly from the game files.

The old list still pulled the stats from the Wiki. The only other option
would be to write our own parser and updated that read through the ggpk,
which is currently what the Wiki does.

brather1ng:
dist-update (actually updates the correct directory now) and all old options (except `/F`) are back. ItemDB is now initialized from all files in the Data/ItemDB/ folder of which GemList.xml is the one that can be updated by the gem loader in UpdateDB.

l0g0sys:
From the point of application speed, it's better to have it in one big 
file, which leads to less I/O and deserialization operations, when you 
need all those information anyway. I think better name would be simply 
Gems.xml in your new ItemDB layout.

There is no point loading all gems as not all gems are supported by app 
so it would be just waste of space. ItemDB API actually doesn't need XML 
at all to return data about gem. XML is there to return more precise 
data, if gem is listed there (for +X gem level mod purposes mostly 
atm.). Back then, in XML were only gems which were taken into account by 
Compute class and I personally tested them (there is no point having 
Trap gems in DB when they are ignored by app).

ItemDB was meant to hold all game item related data. UpdateDB /G will 
naturaly add/update only gem and not touch any other data stored in 
Items.xml. While using UpdateDB without args would update all data in 
Items.xml using all data sources.

ItemDB API supports merging of data from different files, so /M switch 
is natural way to expose this feature. It's not really mandatory to have 
it, just thought it might be useful in the future.

/F switch is "common sense" switch. If you make CLI tool which update 
files, it's natural to have option to have a way to tell tool which file 
you want to work with. While app uses fixed name file (Items.xml), there 
is no harm in having such switch.

I think Ttxman just didn't bother to look into ItemDB API at all, 
otherwise he wouldn't have to make separate parsers and files for 
storage of those affixes.

l0g0sys:
So, if you think it's worth the time to change whole concept of ItemDB API as it will make stuff easier for maintainance and usage, just go ahead :)
If I was going to cleanup this stuff, I'd just extend ItemDB API and UpdateDB to work with affixes stored in Items.xml with parser implemented in UpdateDB.

I've just  found it better to have one file to deal with during development, usage, customization and maintenance (upgrade etc.) opposite to predefined layout of multiple small files with different syntax and some arbitraty logic to handle all those different files.
When all you have to do is process single file (possible XSLT transformations, XML schema validation, and whatever will future bring) in uniform way.

brather1ng:
To hopefully put an end to this discussion:

> From the point of application speed, [...]

Yes, sure it's faster with a single file containing only the necessary gems. But we are speaking about unnoticeable timeframes so I don't really care about performance in this case.

> So, if you think it's worth the time to change whole concept of ItemDB API as it will make stuff easier for maintainance and usage, just go ahead :)

Because Ttxman already made the affix and base item lists, it was easier to keep them independent of ItemDB. I only wrote the parsers and changed the format a bit. If I started from scratch, I probably would have done it differently. I did wonder why he made that stuff completely separate.

> I've just found it better to have one file to deal with [...]

Well, we probably have to agree to disagree on that. I don't like huge files at all :)

SpaceOgre:
Some comments (Only a fast look for now):
* I think it might be more clear to change the "OK" button in the new dialogs to "Close", since this is all it does.
* It seems the expander buttons have "slid" down a bit:
image
* The Reset button will reset the ascendancy tree even if it is unchecked in settings.

EmmittJ:
The third point, doesn't resetting the tree in the current version reset
the ascendancy tree?

brather1ng:
I made the reset button adjustable through the settings as discussed in #324. It can be changed to not reset the ascendancy tree (if it wasn't bugged).

SpaceOgre:
Really nice work so far!

All my "buggs" are fixed. 

Noticed something else now (very minor), I think it might have always been like this as well, but if I choose to change language we open a message box. So far so good, but the focus is not on the "Ok" button in the new window so when I press enter the settings window is closed instead.

brather1ng:
That is caused by the way MahApps dialogs work. They are not really dialogs, they are just overlays that *try* to block interaction with everything else (for example, you can still use features via keybindings) as opposed to the native modal dialogs that reliably block everything.

In this case, the keyboard focus stays on the opened combo box on the dialog below for some reason. I'll see if I can fix it.

SpaceOgre:
Ah, that explains a lot then. Don't spend to much on it if you feel like other things are more important, it is not a big deal after all.

brather1ng:
I know I timed this pretty badly, but I'm pretty much done.

I still have to write something down on the new dialog stuff and adjust the remaining old dialogs. However, that shouldn't affect anything else.

EmmittJ:
Very cool! Again, thanks for working on this so much

brather1ng:
You can now run dist-updateItemImages.bat to load the base item images into your debug directory. So nothing should stop you all from testing it ;) (Don't feel pressured to do that, I still have stuff to do, it just doesn't really matter if I do that before or after this being merged)

EmmittJ:
Hey, sorry I have been rushed with finals. I will take a look at this tomorrow some time.

EmmittJ:
http://pastebin.com/wbMZzLCq

When loading this build I get `Unknown item type: Karui Sceptre`, but this item is in the `Itemlist.xml` file.

EmmittJ:
Also, through no fault of your own GGG has changed the way item images are linked in the skilltree data. This means that re-downloading assets with the new tree data breaks.
`https://p7p4m6s5.ssl.hwcdn.net/` will need to be added to the front of the `/gen/image/` assets

Example:
 * old `https://p7p4m6s5.ssl.hwcdn.net/gen/image/YTozOntpOjA7aToyMjtp/OjI7YToyOntzOjE6InQi/O2k6NDE7czoyOiJzcCI7/ZDowLjM4MzUwMDAwMDAw/MDAwMDAwNzU0OTUxNjU2/NzQ1MTA2NDQ3NDg4MDY5/NTM0MzAxNzU3ODEyNTt9/aToxO3M6MzI6IjY4OWZm/OWE1YWRlZWViZTdlNWQ5/N2JhZTk0ZjcwNjIxIjt9/601c5654c6/Saboteur.png`

* new `\/gen\/image\/YTozOntpOjA7aToyMjtp\/OjI7YToyOntzOjE6InQi\/O2k6NDI7czoyOiJzcCI7\/ZDowLjM4MzUwMDAwMDAw\/MDAwMDAwNzU0OTUxNjU2\/NzQ1MTA2NDQ3NDg4MDY5\/NTM0MzAxNzU3ODEyNTt9\/aToxO3M6MzI6IjFjNTlk\/MDUzYzljNzAxOTQwZWE4\/MWVkYjNkYTIzMGRlIjt9\/b7b1ef47b3\/Trickster.png"},"ClassesSaboteur":{"0.1246":"\/gen\/image\/YTozOntpOjA7aToyMjtp\/OjI7YToyOntzOjE6InQi\/O2k6NDE7czoyOiJzcCI7\/ZDowLjEyNDYwMDAwMDAw\/MDAwMDAwMjQyMDI4NjE5\/MzY4Mjg0MTI1ODEyMzUx\/NzAzNjQzNzk4ODI4MTI1\/O31pOjE7czozMjoiMzIw\/OTNjNWZjZTg4NDU3ZTYy\/MzljMDdkZjA5MDI0ZjIi\/O30,\/46d3fb53f1\/Saboteur.png`

brather1ng:
Fixed the first issue.

I can't reproduce your second issue. For me, the image links are still the old ones. I did fix a bug where the old skill tree file was not removed by redownloading assets, so maybe that fixes your issue.

EmmittJ:
This isn't a major bug, but when dragging weapons out of the inventory they all appear to the the size of a 2 hander (2x4) instead of their actually size. This is only a problem with weapons

brather1ng:
Yeah, happens with shields too. I'll try to fix that, but that shouldn't stop you from merging if you think it's ready (and press the correct button this time :P)

EmmittJ:
Yeah, it looks good. I clicked through pretty much every button and menu. I probably missed something though

brather1ng:
It did work correctly.

EmmittJ:
It was default to the other option which doesn't make any sense at all.

brather1ng:
Weird, there was only one option for me.

EmmittJ:
Hmm, I had squash and merge and commit and merge with squash and merge
being default

PR #327, opened 2016-05-29
Updated Slovak translation and language switching fix.

l0g0sys:
Updated Slovak translation and language switching fix.


SpaceOgre:
Nice, cleaner than my attempt. Good to keep persistentdata and L10n as separated as possible.

l0g0sys:
yep, that's the point :)

Issue #202, opened 2015-08-21
Simple item filter editor


l0g0sys:
Some time ago i wanted to dig into XAML/UI stuff and ended up working on simple item filter editor.
Idea is to make it as simple as it can be, so people doesn't have to know anything about their syntax.
There is nice app Filteration on github, but when i showed it to my brother, he just laught at me. And tbh. i can't be bothered with it either, when i want to change just few visuals or hide some stuff.

Feel free to evaluate from my repo:
https://github.com/l0g0sys/PoESkillTree/releases/download/2.2.2/PoESkillTree-ItemFilter.zip
Made it portable so it won't overwrite anything, except item filters. Be careful, to not overwrite your existing ones using same name.

Filter management window (atm in View / Item Filter Editor, maybe it should be in Tools?):
![itemfiltermain](https://cloud.githubusercontent.com/assets/8587407/9410003/1d1fa5b2-481e-11e5-8a12-4fcdfcd19b2b.jpg)
Window where you create/edit/delete filters. They are stored in AppData.Folder("Filters") and generated into My Games\Path of Exile game folder once saved.

Filter editing window:
![itemfilteredit](https://cloud.githubusercontent.com/assets/8587407/9410157/f1e9d0e2-481e-11e5-89ad-4075dc436d6d.png)
This is simple. Keep checked what you want to see, uncheck what you don't. Disabled rules cannot be hidden due to either "security" reasons (to not make people hide high value stuff by mistake), or there is no meaning in hiding it atm. (Or it could produce wrong file once unchecked :-p)
Rules are grouped to some logical groups like Currency, Recipes (to highlight stuff which can be handed as recipe to verndor), Flask (nonunique), Gems, and few others i found useful atm.
Rules are hardcoded so they can be localized. Specially descriptions, where should be mentioned which items are controled by rule, pretty much copy & paste from Wiki of recipe description.

Color editing window:
![itemeditorcolors](https://cloud.githubusercontent.com/assets/8587407/9410417/5c3366ec-4820-11e5-838a-b3868e1ec528.png)
Set background, text and border color for rule, or whole group of rules. If you want to keep in-game color just leave color at Default.

Still some processing of data before filter file generation is needed.
And think some way to allow hiding of "white trash" and "blue trash", maybe even rares if there are people hiding those? So, you can hide white trash but still show whites for armourer scrap/whetstone recipes, etc.

EmmittJ:
[...]

PR #331, opened 2016-06-02
Simple Item Filter Editor

l0g0sys:
See #202 for more info.


l0g0sys:
To evaluate, just go to my repo, download as zip, double-click on solution and Run.

MLanghof:
Not like it's my business to decide, but why do we need an item filter editor in the passive tree planner? There's *plenty* of those already...

EmmittJ:
I think I would tend to agree with @MLanghof. It is a build planner, and I don't really see how item filters fit within that theme. We may want to expand from being just a build planner? Some sort of multi-use tool?

l0g0sys:
@brather1ng
> Copying a filter works really unintuitively. I only discovered how it worked on accident (enter the new name, select the old filter, click copy). A small input dialog for the new name that is opened via context menu or create copy button press would be much better.

It's all written there, what to do. No need for intuition.
Think, we decided some time ago to avoid context menus, those are least intuitive and not much touch friendly (we still consider touch device support).

> The expected behaviour for double clicking a filter is that it opens the edit window.

Done.

> The check boxes for most rules are disabled.

Yep, to make sure anyone will hide important stuff by mistake. And there are still missing some mechanisms to allow hide more stuff and not screw people.

> Recent colors are not really kept (only in the color picker for that color and only until "Choose Colors" is closed) which makes them pretty pointless.

Yeah, there are issues with this due to color names. So, for now it's on TODO list.

> If you do not want to use the Extended WPF Toolkit as NuGet package for whatever reason, you should probably put a license notice on the style file. I would suggest to move the style file and the dll together with the license into a subfolder of Contrib.

The reason is written right there in README.txt. Contrib folder is for use of 3rd-party source code directly, not for DLLs. I've made Libs forlder for 3rd-party DLLs (the rest of non-NuGet libs should be moved there with time).

> Imo, all the new files should be moved into an ItemFilter subfolder (like you've done with the files in Model but with ItemFilter as the outer folder) as they don't have anything to do with the rest of the program (except the stuff in Utils maybe). Moving them into a new project would be better, but that's probably not possible because of cyclic dependencies.

Yeah, noticed how you refactored TreeGenerator. Refactored code as well to match that layout.

@MLanghof
If you can point me to one of those editors, which are really simple to handle and doesn't require any knowledge of item filter mechanics, I'd be really glad to check it out. All I found so far was just full item filter syntax made into UI. Not somthing common people can use.

Regarding need or don't need, that's philosophical question. I was thinking about it as well, that we add this feature and now suddenly 10 people will come up with some other non-build related features. But I doubt it will happen. We can draw a line, if you want, and say only this and nothing else. We already crossed the line from simple “click a node” application, so now it depends how far we want to go. You can look at it as extra feature to help people, it's not like it'll redefine whole purpose of application.

But I can come up with some build related stuff, if you want :p
Like it could create you item filter for your build, so you could have highlighted item drops for your build (weapons, armor types, uniques, …).

From my point of view, it's no brainer to exploit this application with all its foundations to accomplish such simple tool, than work on standalone application from scratch with my limited free time.

If only you brought this question earlier. I've submitted issue (#202) about it like almost year ago and only @EmmittJ commented.

brather1ng:
You forgot to change the namespace of the views, doesn't compile atm.

> It's all written there, what to do. No need for intuition.

It's not really clear to me from what's written there.

> Yep, to make sure anyone will hide important stuff by mistake. And there are still missing some mechanisms to allow hide more stuff and not screw people.

Hm, I get why they are disabled but it'll probably confuse people if they don't know the reason. I thought it was a bug at first. I don't have a better idea, though. Also some of them should be hideable (scraps, whetstones and baubles as currency and recipes, 5 links, jewels, and the chance bases)

> The reason is written right there in README.txt. Contrib folder is for use of 3rd-party source code directly, not for DLLs. I've made Libs forlder for 3rd-party DLLs (the rest of non-NuGet libs should be moved there with time).

I'm fine with where you put them now. (I don't agree with you on the reason for non-NuGet, but I don't mind an additional separate dll so I won't start a discussion on that)

On the discussion as to whether this should be included: I don't mind it since it is just a menu entry in Tools. However, we should not advertise it as a big new feature and make clear that it's just a small and simple tool that doesn't change the main purpose of the application.

l0g0sys:
> You forgot to change the namespace of the views, doesn't compile atm.

It did. Just had to Clean solution first. I'd not commit someting which doesn't compile. I always do Clean and Build solution before commit ;)
Anyway fixed.

> Hm, I get why they are disabled but it'll probably confuse people if they don't know the reason. I thought it was a bug at first. I don't have a better idea, though. Also some of them should be hideable (scraps, whetstones and baubles as currency and recipes, 5 links, jewels, and the chance bases)

I've got it tested since August 2015 by my guildies and they didn't complain at all about restrictions. And I'll tell you, they are far from being PC gurus :) They are happy they can play with colors and that's it.

Yeah, some lower currency could be enabled for hide. Though, recipes and similar complex rules probably not due to overlapping sets of matched items. For now, I enabled only those rules which were thoroughly tested (few months) and don't create filter which would hide something of interest (anything checked).

The purpose of the tool is to color stuff yourself on the ground with easy. The hiding feature to reduce clutter is just "nice to have" and with lot of work ahead.
Ultimate goal is to include shard recipes, so you will be pretty much able to hide all trash on ground, while keeping interesting stuff visible.

But currently, my top priority is to add support for new and neglected mods since patch 2.1 or something. So, I'd like to have this issue closed ASAP, while I still have bit more of free time for few weeks :)

brather1ng:
Looks good to me now. Though, someone with more lootfilter experience should probably test it too.

> So, I'd like to have this issue closed ASAP, while I still have bit more of free time for few weeks :)

There's nothing stopping you from working on that while this is open ;)

l0g0sys:
Yeah, except extra work for maintaining branch and own releases for my 
guildies. Guess I'll have to live with that for a while.

EmmittJ:
Sorry I haven't been more active here (finals week next week) I will take a
look at it this afternoon.

l0g0sys:
Np, just thinking about build for Prophecy :p

And you have exams right at league launch, bah.

EmmittJ:
Yep, I do. It is pretty annoying >.> I should be able to look at it before
the league launches, but that is pretty much it.

EmmittJ:
The item filter works as far as I can see. If we missed something then
people will let us know.

EmmittJ:
so what is everyone's vote on merging this?

MLanghof:
Tentative no (for what little my vote is worth).

> If you can point me to one of those editors, which are really simple to handle and doesn't require any knowledge of item filter mechanics, I'd be really glad to check it out. All I found so far was just full item filter syntax made into UI. Not somthing common people can use.

Leaving the inevitable shortcomings of such a simplified filter aside (which are probably the reason why few/none of them exist yet), I can sort of see the need for such an application. On the other hand, the people that would hesitate to touch any existing loot filter builder will be satisfied with the loot filters other people have made. And if they wanted to change something, if it's just style/appearance then Filtration (or the other mechanisms the authors provide) is certainly the tool for that, and if it goes beyond that, I don't see your tool providing that.

TL;DR: How would a filter from your editor be superior to any existing ones?

> Regarding need or don't need, that's philosophical question. I was thinking about it as well, that we add this feature and now suddenly 10 people will come up with some other non-build related features. But I doubt it will happen. We can draw a line, if you want, and say only this and nothing else. We already crossed the line from simple “click a node” application, so now it depends how far we want to go. You can look at it as extra feature to help people, it's not like it'll redefine whole purpose of application.

Sorry, but there's no argument in there. I'm reading this as "I know we shouldn't do it, but I guess/hope it wouldn't be all that bad if we did".

> But I can come up with some build related stuff, if you want :p
Like it could create you item filter for your build, so you could have highlighted item drops for your build (weapons, armor types, uniques, …).

That would be something innovative indeed, but I'm afraid we're not really at a point where we can infer that kind of info though.

> From my point of view, it's no brainer to exploit this application with all its foundations to accomplish such simple tool, than work on standalone application from scratch with my limited free time.

I totally understand that reasoning. I also see why you wouldn't want to make it a separate application ("yet another program I have to install for PoE"). Integrating the filter into the passive planner is the best option for *you*, but the question is if it's also the best option for the *passive planner*.

l0g0sys:
It's superior in its simplicity. That's it.

Guess you don't know many people who have no idea how filters works, and 
just want some common stuff to not miss.

I know how filters works, and I still not gonna use Filteration. Too 
much hassle for simple stuff i want my filter to do.

Not to mention that you can screw yourself if you don't pay attention to 
what you're doing in such tools. Like you can mess order of rules.

It's like asking why the hell people are using Prolog when they can 
write everything themself in C ;)

brather1ng:
My vote: I don't really mind whether this is merged or not.

I don't think many people will use it and it has nothing to do with tree/build planning, but it just sits in the tool menu not doing anything for people that don't want to use it. Plus, the code is mostly separated from the main code, so I don't see this becoming a big maintenance issue.

Well, the popup code has to be adjusted once my pull request is merged and the dialogs will look different than those of the main program, unless they are adjusted.

l0g0sys:
Well, maybe you could all that async stuff wrap into Popup class 
methods, so we have easy modal-like API where needed and still keep 
async stuff at work?

brather1ng:
They are called nearly identically to the Popup class. The only difference being that they are extension methods on MetroWindow and need to be awaited. So adjusting that is not really difficult as in your code the MetroWindow instance is always available from where Popup is used. It's done in a minute or so.

Also, the async stuff can't be hidden, async has to be propagated through the code up to the calling event handler (at least for code that does something with the UI).

l0g0sys:
Updated async stuff and added new items to filter.
Guess it's ready for a merge.

EmmittJ:
Looks good, I think we can go ahead and merge.

brather1ng:
@l0g0sys Whatever you have done with AffixList and ItemList, I don't think that should stay as it is. (AffixList starts with the HTML source code of a GitHub webpage and ItemList also has 25k additions).

EmmittJ:
I just saw this. @l0g0sys you'll have open the pull request when the problem above us fixed.

PR #345, opened 2016-08-02
Expand information in GemList

rasmuskl:
Changes to GemList.xml:

* Fixes parsing of the info box (Gamepedia switched from div to span)
* Adds Cast Speed, Critical Strike Chance, Cooldown Time and Damage Effectiveness as attributes
* Adds Tags attribute containing skill gem tags

### DIFF
--- a/UpdateDB/DataLoading/Gems/GamepediaReader.cs
+++ b/UpdateDB/DataLoading/Gems/GamepediaReader.cs
@@ -280,15 +280,6 @@ namespace UpdateDB.DataLoading.Gems
                     });
                 }
 
-                if (textDefault.InnerHtml.Contains("Radius") && Regex.IsMatch(value, @"^\d+$"))
-                {
-                    attributes.Add(new Attribute
-                    {
-                        Name = "Radius: #",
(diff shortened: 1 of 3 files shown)


EmmittJ:
I am curious as to how radius is inferred by the current data?

brather1ng:
Looks good to me. I don't really like the tags being a combined single string, but that seems to be the only way to get it into a Xml attribute (well, a Flags Enum is another option, but that's a lot of effort for something that's not yet used).

@EmmittJ Did you missread "interfered" as "inferred"? Radius is not used anywhere atm.

rasmuskl:
I originally had Radius in the Gamepedia changes, but the `TestItemDB.TestGems` test failed:

image

I re-added the change now though and debugged the test. It turns out it happend because Elemental Profileration Support lists Radius twice, it wasn't interferring with the old data, but it was already being parsed.

image

I suspect this is because the Radius used to vary by level on Elemental Proliferation Support.

I've re-added the Radius parsing and added a check to skip duplicates.

EmmittJ:
@brather1ng yes...and the flags won't be a big deal until they are used like you said.  
@rasmuskl looks good to me overall. I will go ahead and submit this request

PR #357, opened 2016-09-02
Revert "Simple Item Filter Editor"

EmmittJ:
Reverts EmmittJ/PoESkillTree#331


EmmittJ:
There was some problems that I didn't see when merging the item filter code

PR #360, opened 2016-09-02
Changed AppId to reflect new product release to be installed alongside of previous release

l0g0sys:


brather1ng:
Just a suggestion: use feature branches instead of doing everything on your master branch. That way, you can just branch off of the main repo's master and delete the branch when it is merged.

Also, what happens if people update from the released version to this one (which has the same program name but different AppId)?

l0g0sys:
It will pre-fill destination folder of previous version in installer. So 
if people don't notice it, it will copy new files there.

Tho roaming profile will be created new as it's separate functionality 
from Windows registry handling based on AppId.

brather1ng:
Ok, sounds like there shouldn't be any problems, other than maybe two uninstallation entries for basically the same thing.

l0g0sys:
Current 2.3.0 tag should be removed. New release made and then update 
release with new zip/exe and new 2.3.0 tag.

Not sure if it's possible to leave release witgout tag tho. Or it has to 
be deleted and recreated with new tag.

brather1ng:
Can't we just release it as 2.3.1 and state in 2.3.0 that it shouldn't be downloaded (or remove 2.3.0 if that's possible)?

If we release another 2.3.0, I don't think the update function will recognize it as an update.

l0g0sys:
Actually, there will be only one uninstall entry. It won't possible 
uninstall previous installation anymore.

l0g0sys:
I'd rather fix current 2.3.0 release.

So those who didn't update yet will get correct installation with 
separate uninstall entry.

Having it fixed by 2.3.1 isn't good idea, 'cause it would make those 
already having 2.3.0 installed, get another uninstall entry, but using 
same AppData folder as 2.3.0 installation. That would be bit of mess.

brather1ng:
It will eventually be fixed in a version that the ones that currently have 2.3.0 will get. So the mess *will* be created anyway.

And fixing 2.3.0 without a new release will mean everyone currently on 2.3.0 will have a harder time to get the bug fixes that are contained in it.

l0g0sys:
Heh, yeah. There is no way out of it :) Guess making 2.3.1 will do it.

l0g0sys:
I'll have to look into release.xml if i can script some AppId check when 
major/minor version is being changed to avoid this in future.

brather1ng:
Another option: release 2.3.1 with a new name (like "Atlas of Worlds Fix" or something). Wouldn't clash with anything old, but would require manual build copying.

> I'll have to look into release.xml if i can script some AppId check when major/minor version is being changed to avoid this in future.

Better to do an AppId check if program name changed (and vice versa probably), if that is possible. Wouldn't lock us to only increase the minor version with game expansions.

l0g0sys:
Changing ProductName won't help anything. Just would result in new 
AppData folder on launch which installation/update process doesn't care 
about anyway.

The issue now is that people got Ascendancy .exe file overwritten by 
Atlas of Worlds .exe. And we cannot do anything about it.

So... making just new release is enough to fix it so people will have it 
installed into new suggested Atlas of Worlds folder with new uninstall 
entry.

And then they can just redownload last 2.2.x version and reinstall it 
into Ascendancy folder to have previous version running alongside of 
Atlas of Worlds.

Also, now i think rising version to 2.4.0 wouldn't be bad to have 
version aligned with GGG's 2.4 content patch :)

brather1ng:
> Changing ProductName won't help anything. Just would result in new AppData folder on launch which installation/update process doesn't care about anyway.

Ah, that's what you meant with the mess. Yeah, then a new ProductName wouldn't fix anything.

> Also, now i think rising version to 2.4.0 wouldn't be bad to have version aligned with GGG's 2.4 content patch :)

No, we don't want to do that. People then would think we would always use their version scheme and complain even more :)

EmmittJ:
I forgot to change the appid in the update process.

I don't think we should raise the minor version number. It doesn't make
sense in a development world. Why would we update the program to match GGG
this has never been a thing in the past?

l0g0sys:
I always thought it would be easier to spot compatibility of app with 
game by having major.minor aligned.

But forget i mentioned that :) Was just a brainfart idea long time ago. 
It just came back to me when i noticed that now we have finally 
opportunity to do it if we choosed so as GGG jumped to 2.4.

l0g0sys:
Oh no, not a brainfart idea. Damn, i've got side tracked.

There might be a problem with making 2.3.1 release but with changed 
AppId. It will result for those already having 2.3.0 installed and doing 
"Check for updates" in silent installation into some folder. Really 
don't know into which folder it would install. If release is an update 
(last digit of version string is changed), then "/SP- /SILENT /NOICONS 
/LANG=" + GetSetupLanguage() is appened to installer exe on launch of 
update process. Thus no dialogs are shown!

Think we will have to do 2.4.0 release in the end to have all existing 
2.3.0 installations detect non-silent installation of new product :(

EmmittJ:
What would be the affects of a silent update with a changed app id? The app data folder should already be in place and it would create a new install directory and program in the programs list, correct?

l0g0sys:
No idea what will happen. In silent install it should take destination 
folder from registry and install it there. But if AppId was changed, 
there is no registry entry yet. So it's kind of conflict here. Think it 
should be tested first. Change last digit, do dist-release and then 
launch built installer with /SP- /SILENT /NOICONS /LANG=en_US arguments 
to simulate auto-update, and see what happens :)

PR #362, opened 2016-09-08
New Build Saving Structure

### LINKED ISSUE #14: Progression support for builds
Many build guides use a level progression, so I think it would be nice if we could some how add this to a build. Something like this:

**Saved Builds**
* Build1
  - lvl1-20
  - lvl21-40
* Build2

Just a fast concept :)

### LINKED ISSUE #192: New Build Saving Structure
I would like to open up a discussion about possibly a new build saving format that doesn't keep them all in one file. 

I believe persistent data should be thing related to settings for the app (maybe a rename), and we have a completely different setup for builds.

I suggested that we do a folder set up but as @l0g0sys pointed out there are a couple problems with this structure:
1) Naming: Build name would have to be separate information from file name. Unless build name would be more strict about allowed characters to make 1:1 build name to file name mapping. And question is what role subfolders would be playing in build naming.
2) Sorting: There wouldn't be any manual sorting, maybe just automatic sorting based on name or creation/modification time. Using some toggle buttons.
3) Filtering/Searching: This would require to traverse folder structures and parse each build file to filter based on character class or search in notes. Unless whole structure of build files would be read on launch and kept in memory during runtime.

### LINKED ISSUE #338: Request: Save tagged nodes with saved build and sharable tagged nodes
Currently tagged nodes are not saved between runs. If you close the application then you lose whatever tagged nodes you had setup.

In addition tagged nodes are not linked to builds. If you are switching between builds and want to use the feature you need to open multiple instances of the program to accommodate them.

What I would like to see is the ability to save tagged nodes on a per build basis and have them persist across sessions. It would also be nice to be able to share tagged nodes for assistance and collaboration.

brather1ng:
Builds are saved as separate files in a folder structure. I rewrote pretty much all build related code.

**Bigger changes:**

- Split the control responsible for build management and UI out of MainWindow and improved it.
- Rewrote PersistentData de-/serialization. Code not directly related to serialization will only use IPersistentData. The file is deserialized by different classed based on the version it was created in so old builds can converted.
- Builds are saved in a folder structure:
  - *.pbuild files are the builds themselves (serialized PoEBuild instances)
  - There is one .buildfolder file per folder. This file is optional (there is not much lost when you delete it) and mostly contains the sort order of the builds in that folder. Builds in that list that don't exist are simply ignored and builds not in that list are added at the end.
  - These files contain a version (currently 1.0). The program won't open builds of a newer version and when there are new build versions converters for old versions can easily be added.
  - File names are URL encoded to allow arbitrary names.
  - Folder nesting is currently limited to 4 (including the root) to limit scroll bar appearances in the UI.
- Each build has a dirty state that is set when it is changed. They must be saved explicitly (folders are automatically saved to make handling of hierarchy name changes easier, they don't contain much information anyway). When closing with unsaved changes, the user is asked if he wants to save them or not (or cancel). You can discard changes made through the context menu of builds.
- The currently opened build is the same as the one shown as currently open in the build list (previously they were two separate instances with the one in the build list only being updated on save).
- The context menu is heavily used in the build list as those options couldn't all fit into the UI as buttons.
- Export/Import:
  - The build folder is monitored for changes. When it changes outside of the program (technically, that can't be checked, so changes made outside simultaneously to changes inside will not be recognized), the user is asked if he wants to discard changes and reload or not. If he wants to save some changes before reloading, there is a reload option in the menu. Doing this for individual files would require much more work and I don't think there will be many cases were that would be useful.
  - There is a menu link for opening the build folder.
  - When installing, file associations to *.pbuild are created. This can't be disabled in the installer currently, I'm not sure if it should be an option. If it would be an option, I don't know what the behaviour should be when this is installed through a silent upgrade. If you double click a file in your build folder, the program opens with that file as current build. If you double click a file outside of your build folder, the program loads the build from the build folder normally but asks you for a name under which the new build should be imported (the original file will not be changed).
- If the build folder is not set when starting the program (on new installation or after updating), the user is asked for its location. If there is an old PersistentData file, the folder must be empty to not require handling of its contents and the builds from the old file are added to it. If there is none or a current one and the folder contains builds, the builds from the selected folder are loaded. The folder can also be changed through the settings.
- Tagged nodes and Tree Generator settings are saved as part of a build. Builds have a property for arbitrary JSON to not clutter them with all the Tree Generator stuff.

For smaller changes, see the commits themselves.

Example build folder to play around with: [Builds.zip](https://github.com/EmmittJ/PoESkillTree/files/461755/Builds.zip) (just extract it somewhere and set the build folder path to it)

If someone wants an installer to test the import stuff, I can create one, too.

Because of the amount of changes and file system interactions, I would suggest releasing this as a pre-release first. So, if we want to do a bug fix release, that should be done before merging this.

Closes #338
Closes #192
Closes #14 (I think this is all the progression support we will ever get)

**Screenshot:**
screenshot


EmmittJ:
Not to derail this quickly, but do you think that this should be a major
version jump? The amount of changes to base behavior is pretty high.

brather1ng:
I don't really care much. It would be a pretty big update, so that would make sense.

EmmittJ:
OK, I will take a look and play around with this once I get back from work.

EmmittJ:
Doesn't look like you can paste outside of a folder (to root)?

Moving a build to root then attempting to move anything will prompt the user with the message that the Builds have changed. This seems like odd behavior. Maybe root changes aren't tracked internally? 

Other than these I don't see anything wrong.

SpaceOgre:
Except Emmitts findings I have some small suggestions
* The folders kinda gets lost in the list, especially when empty. Maybe increase there text size a bit? or have a folder icon or something.
* It would be nice to have the ability to expand/collapse all folders. Some users have a lot of builds. This could be added to the View menu.

Great job!

SpaceOgre:
One more thing, the build list filter drop down. Would it be possible to be able to select the group head instead of having to list the base classes again?

brather1ng:
> Doesn't look like you can paste outside of a folder (to root)?

Yeah, the root folder is hidden, so its context menu is not avaiable. Maybe pasting on builds should be enabled and paste to the parent folder?

> One more thing, the build list filter drop down. Would it be possible to be able to select the group head instead of having to list the base classes again?

I don't think that's possible. The group header doesn't have any data representation and is only a visual thing. But I'll try to find a solution, maybe it is possible.

I'll fix the other issues.

SpaceOgre:
> I don't think that's possible. The group header doesn't have any data representation and is only a visual thing. But I'll try to find a solution, maybe it is possible.

Okay, it is not a big thing so don't no need to spend to much time on it. But it would be nice since the list would be shorter.

brather1ng:
> Doesn't look like you can paste outside of a folder (to root)?

You can now paste with a build selected (to the parent folder of the build). This requires a build in the root folder, but I think that's not a big problem, there's always the drag & drop option.

> Moving a build to root then attempting to move anything will prompt the user with the message that the Builds have changed.

I was able to reproduce this once in many tries but don't know what causes it. I have doubled the time window for which file system changes are suppressed after changes by the program, hopefully that's enough to fix this. (it's 200ms now, no one should be able to do manual changes to the build folder in that time frame)

brather1ng:
Group headers in the class filter box can now be selected (they aren't group headers anymore, just restyled entries that look like group headers).

Folders in the build list have a icon. As a side note, I have added the `MahApps.Metro.IconPacks.Modern` NuGet package. Icons from ModernUIIcons can be easily used now:

`<iconPacks:PackIconModern Kind="Folder" />` (use `Height` and `Width` for changing their size from the default 16x16)

SpaceOgre:
Looks good. I think it is ready to merge.

Nice with the icon pack :)

EmmittJ:
http://imgur.com/a/3GtOK

This is what I was talking about when the build message would pop up. You move the build (or folder) to root, and it freaks out. A 2 second delay seems to be about the perfect amount. This gives the file system and the application to be in sync with each other rather than one being much faster than the other.

brather1ng:
Well, that doesn't happen for me at all. There are no changes made when only dragging a build, does it only happen when you drag a build after moving one to root?

Have you tested it with a 2 second delay? I think anything higher than 2 seconds would be too much as that would be enough time to do manual changes.

SpaceOgre:
I'm not getting this either. Maybe harddrive vs SSD or just harddrive read/write speed?

brather1ng:
Neither getting this on my SSD nor on my harddrive. However, I got that change message when switching the build save directory to a relatively filled one on my harddrive.

So, increasing the delay to 1 second seems like a good idea. That should still be unproblematic with manual changes and be high enough for slower drives or drives being occupied by something else.

EmmittJ:
I tested with 1 second, and this weird interaction happens. 2 seconds was
much better. It looks like it only happens when I put things into root and
attempt to do anything else.

brather1ng:
Increased to 2 s now. Hopefully people aren't that fast.

EmmittJ:
Honestly, I am not sure what causes it, but I had a clean copy locally. My build folder was generated and in the debug/bin folder. It is pretty strange.

I think 2 seconds is a good timer. I was messing with builds outside of the folder, and it didn't seem to have a delay even when the program was directly in the background.

After this, I think we can merge. I didn't see any other issues when messing around with it.

EmmittJ:
@brather1ng you think this is ready to merge?

brather1ng:
I'm done, so yeah.

EmmittJ:
good stuff!

Issue #369, opened 2016-09-11
Skill Tree Generator Broken


Shaod:
I know you're super busy, and I doubt many people use this feature, but the skill tree generator is broken in several ways. I'm sure it's tempting to remove it, but for those of us that use it, it's awesome to find optimal and overlooked pathing routes.

Firstly, and probably most importantly, it returns invalid trees. It seems to do this the majority of times, at least if you add multiple points to path to.

![poeskilltree-generator2](https://cloud.githubusercontent.com/assets/22132367/18418068/3da2e656-7837-11e6-9431-f33fdcb3ae05.png)

Secondly, it doesn't respect the 'current tree as initial solution' option.

Thirdly, if you try to open the generator after you've already generated a skill it says that the window is already open, preventing you from running it again or tweaking results / weights.

![poeskilltree-generator1](https://cloud.githubusercontent.com/assets/22132367/18418078/6f03e880-7837-11e6-98b2-b20492faae41.png)

brather1ng:
1. Are you using any attribute or pseudo attribute constraints? If no, this is a known bug but not really relevant because using "Skill Tagged Nodes" does exactly the same thing and doesn't have that bug.
2. I will look into that.
3. That bug is fixed, the fix is just not yet released.

Shaod:
For 1, no I wasn't. Good to know there's an alternative, and thanks for the quick response.

MLanghof:
> I'm sure it's tempting to remove it
Not gonna happen, both brather1ing and me joined this project specifically to create it :)

brather1ng:
For 2, it does work from what I can see. It may not work in combination with 1, as those invalid tree may influence it.

As 1 is neither easy to fix nor really relevant, I'll close this. It has to be fixed at some point anyway, as it is caused by the interaction of the reductions (that are currently not used when constraints are present) and the final hill climbing step (which is not used when no constraints are present).

If I'm mistaken and you do find a situation where 2 occurs, feel free to reopen this.

PR #372, opened 2016-09-15
Improve Tree Generator UI

brather1ng:
These are mainly changes that should make the Tree Generator easier to understand and find, no functionality changes.

- Move "Skill Tagged Nodes" from menu to the generator settings window.
The new "Tagged Nodes" tab contains some explanation as it doesn't
need settings. Settings that don't affect it are disabled when that tab
is selected.
- Change the generator settings window to be fixed width. No more
enlarging when long attribute names are added as constraints.
- Add explanation tooltips to many of the settings.
- Move "Nodes" sub menu to a new top level menu.

Doing this as a Pull Request to give you the chance to comment on the usefulness (mainly @MLanghof as you actually use this a lot).

Also, the "Tagged Nodes" tab still has space for more explanation text. I'd appreciate suggestions on what might be useful to know as a user.


EmmittJ:
I guess I don't mind the change, but this seems to add another window for no real reason? I am kind of confused why it has been taken from 2 clicks to 3 + a new window.

brather1ng:
The window and everything was already there. I just moved "Skill Tagged Nodes" to it. That was the plan since this was initially added (I had both at first). (also, it's one click less now, as "Nodes" is now a top level menu entry)

I could just add the "Skill Tagged Nodes" menu entry back. But it should also be available from the generator window to make the connection between them clearer and have some space for explanation (see e.g. #369).

EmmittJ:
Ah, ok.  I think it is fine, and the only reason to add it back is if it takes you to the new tab

MLanghof:
I recently set up the new VS version on my PC but had to ditch it (the PC) since I moved to a different country. I'll see to getting the 2015 version on this tablet (multi-GB downloads via wireless, yay), so bear with me for a while. I'll make sure to check it out and let you know, but I suppose you can merge this anyway - if there's anything that should be amended then that shouldn't be a problem.

brather1ng:
Thanks, I'll upload a compiled version tomorrow so you don't have to download VS.

MLanghof:
Well, installing it actually went faster (and easier) than I would've expected. I think I can already take a look right now.

MLanghof:
Here's some things that caught my eye:

- "Reset Settings" puts you back in the "Tagged Nodes" tab, I doubt that's what one would intend to happen.

- I would reword "do not expect optimal solutions from the 'Advanced' generator" to something like "the resulting solutions may deviate slightly from the optimum". I know that "not optimal" technically means exactly that, but it leaves a different impression to the average reader. And in any case, the input constraints are only an abstraction of what the user wants, so the nature of our weighting function(s) likely already introduces more significant deviations from what would be the optimal solution the user is looking for.

- I've never had a use for the checkboxes on the right - "Only search in current tree" is an arbitrary restriction that I can't imagine a workflow for (well, if there was a way to skill all nodes in an area of the tree then maybe that would work, but I don't see anybody clicking 200 nodes manually to limit the generator to that area), and I'm not under the impression the algorithm really needs an initial tree in virtually any case (it's very flexible in the early generations anyway). In either case I'd be more worried about shoehorning the GA into a local optimum than preventing a complete failure (which we've never even observed afaik), and the second checkbox certainly is redundant for the "Tagged Nodes" solver (it really doesn't need hints).

Given the above, I'd imagine throwing the level etc. edits and the "Reset Settings" button (and the two checkboxes on the right if you want to keep them) into the Advanced tab would make the UI more intuitive (also eliminating the need for the corresponding paragraph in the mini-help).
So in the end the only controls outside the tabs would be the include/exclude checkboxes and the run button. The "Tagged Nodes" tab would have nothing but the explanatory text, and everything else goes in the "Advanced Tab" (because it only has an effect there\*).

\*Well, if we keep "only search in current tree" then I guess it would/should affect both generators. Would like to hear your opinion.

brather1ng:
"Only search in current tree" was meant as a possibility to create leveling trees, but I have no idea if anyone would use it for that. It's probably simpler to create leveling trees manually. I'll just remove both of them for now, makes both UI and code simpler.

I also agree with your other points. Thanks for checking it out!

brather1ng:
All done and also extended the explanation text.

MLanghof:
Alright, looks good. I feel like the settings tab looks slightly funky (the space between the group boxes), but that's not major.

I also noticed that "include Check-tagged nodes" has to always be enabled for the "Tagged Nodes" tab, otherwise it doesn't make any sense. So that one should probably also be ferried over to the Advanced tab. The last remaining checkbox does have its uses, but it will feel slightly out of place. I don't really know what we should do with it...

brather1ng:
Yeah, well it somewhat boils done to having the same as before plus help text. I do like this more because the relationship between the two generators is obvious this way. As opposed to people unnecessarily running the advanced one as before.

I would like to keep the two checkboxes avaiable for both, mostly so "Tagged Nodes" doesn't look completely deserted.

Regarding it getting in the way, I did like this more than the three clicks away menu entry when I had to use it a ton for testing. Also, you can somewhat put it out of the way by moving it to the bottom of the window until only the run button is reachable.

What I could do is adding keyboard shortcuts for both generators so they can be used without the window being open.

> I feel like the settings tab looks slightly funky (the space between the group boxes), but that's not major.

Yeah, I experimented with that but didn't came up with a good solution.

MLanghof:
> Also, you can somewhat put it out of the way by moving it to the bottom of the window until only the run button is reachable

That was also the workaround I came up with, but surely that's not what UI design should aim for? :P
But yeah, a shortcut would solve that. Then again, if it still requires the Skill Tree Generator window to be open, it's not that much of a gain (because you'd still have to cram the window somewhere every time you work with it).

I'm not sold on keeping "include Check-tagged nodes" available for both generators. I guess it could be disabled while on "Tagged Nodes", but then it'd have to be "frozen" in checked state (undoing the choice the user had for the advanced tab) or be ignored (which would be confusing) or frozen in the wrong state (which would be very bad).

I can get behind the value of juxtaposing both generators for the reasons you mentioned. It just brings some workflow oddities with it.

brather1ng:
This should've been the last commit now.

Both generators have menu entries (that make their relation to the "Skill Tree Generator" window clear) and keyboard shortcuts now. Those don't open the generator window but use the settings from it.

There are no general settings anymore. "Tagged Nodes" has just the Cross-tagged nodes checkbox. Run buttons are also in the tabs.

New users should see the explanation text as the direct menu entries aren't clear without context on purpose. Others don't need to ever see it again as the opened tab is saved and "Tagged Nodes" will probably be used only through the direct menu entry by them.

MLanghof:
Wow, this is awesome! I didn't even think about making the functionalities accessible through both ways, but it elegantly solves both the "wtf does this do?" side and the "don't clutter my UI!" one. Brilliant! :D

brather1ng:
Thanks for all your input, wouldn't have gotten there without it!

Issue #380, opened 2016-09-27
Attack speed "more" qualifier calculated incorrectly


jrobinson3k1:
Currently, "more attack speed" quantities are being added together before being combined with "increased attack speed". "more" qualifiers should be multiplied together, not summed.

Source: http://pathofexile.gamepedia.com/Stat#Flat.2C_additive_and_multiplicative_stats

Example:
76% increased attack speed (from tree)
10% more attack speed (from DW)
93% more attack speed (L19 Multistrike)

Current: (1 + 0.76) * (1 + 1.03) = 3.5728
Correct: (1 + 0.76) * (1 + 0.10) * (1 + 0.93) = 3.73648

Relevant source file: Compute.cs, Lines 627-647

I'm not sure if this is globally miscalculated in the app for other types. I have not checked.

jrobinson3k1:
I'm working on a PR, but I'm not very familiar with the source code yet and don't primarily code in C#, so it'll probably be very rough around the edges.

EmmittJ:
Should be a pretty simple change. You add 1 to the value and multiply then
together. Currently they are summed, so you would just need to change the
one line. (Two in this case because less is also calculated wrongly)

PR #381, opened 2016-09-27
Fixes incorrect "more" calculation for attack speed #380

jrobinson3k1:
I needed to convert the percentage values from whole numbers to their decimal representation in order for the calculation to be correct.

### DIFF
--- a/WPFSKillTree/SkillTreeFiles/Compute.cs
+++ b/WPFSKillTree/SkillTreeFiles/Compute.cs
@@ -624,23 +624,20 @@ namespace POESKillTree.SkillTreeFiles
                     if (incAS != 0)
                         APS = IncreaseValueByPercentage(APS, incAS);
 
-                    float moreAS = 0;
+                    float moreAS = 1;
                     if (attrs.ContainsKey("#% more Attack Speed"))
-                        moreAS += attrs["#% more Attack Speed"][0];
+                        moreAS *= attrs["#% more Attack Speed"][0] / 100 + 1;
                     if (attrs.ContainsKey("#% less Attack Speed"))
-                        moreAS -= attrs["#% less Attack Speed"][0];
+                        moreAS /= attrs["#% less Attack Speed"][0] / 100 + 1;
                     foreach (var attr in attrs.Matches(ReMoreAttackSpeedType))
                     {
                         Match m = ReMoreAttackSpeedType.Match(attr.Key);
                         if (m.Success)
-                        {
                             // XXX: Not sure there are any mods with WeaponType here (Melee string in mod is DamageForm now, maybe Unarmed should be form as well).
-                            if (Weapon.Types.ContainsKey(m.Groups[2].Value) && Nature.Is(Weapon.Types[m.Groups[2].Value]))
-                                moreAS += m.Groups[1].Value == "more" ? attr.Value[0] : -attr.Value[0];
-                            else if (DamageNature.Forms.ContainsKey(m.Groups[2].Value) && Nature.Is(DamageNature.Forms[m.Groups[2].Value]))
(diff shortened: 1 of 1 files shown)


EmmittJ:
I think I stated in the issue something wrong, sorry. I believe the less attack speed here is wrong since you need to multiple the product with the more attack speed product.

jrobinson3k1:
@EmmittJ should be good now

EmmittJ:
What happens when you have an odd amount of less attack speed? I believe
you will get on overall negative attack speed?

jrobinson3k1:
Yeah, I think there's a deeper problem with how "more" and "less" values are stored in the attrs map. I'm not sure where those values are obtained in source, but I assume the collective value is just added together for the "#% less Attack Speed" and "#% more Attack Speed" value?

jrobinson3k1:
The "more" and "less" values are calculated incorrect for every value that I'm seeing, so maybe it's best to discard the PR and create a new issue to address the global problem and fix it all in one fell swoop.

EmmittJ:
Yeah, I can correct it. The two values need to be calculated separately
(products) then the product of those is the final result.

WPFSKillTree/SkillTreeFiles/Compute.cs
```diff
@@ -624,23 +624,20 @@ public void AttackSpeed(AttackSkill skill, AttributeSet attrs)
...
                         APS = IncreaseValueByPercentage(APS, incAS);
 
-                    float moreAS = 0;
+                    float moreAS = 1;
                     if (attrs.ContainsKey("#% more Attack Speed"))
-                        moreAS += attrs["#% more Attack Speed"][0];
+                        moreAS *= attrs["#% more Attack Speed"][0] / 100 + 1;
                     if (attrs.ContainsKey("#% less Attack Speed"))
-                        moreAS -= attrs["#% less Attack Speed"][0];
+                        moreAS /= attrs["#% less Attack Speed"][0] / 100 + 1;
```
MLanghof: I'm pretty sure this should be `moreAS *= 1 - attrs[...]/100;`.
100% less attack speed means you never attack, not that you attack half as often.

EmmittJ:
Yep

WPFSKillTree/SkillTreeFiles/Compute.cs
```diff
@@ -624,23 +624,20 @@ public void AttackSpeed(AttackSkill skill, AttributeSet attrs)
...
                         if (m.Success)
-                        {
                             // XXX: Not sure there are any mods with WeaponType here (Melee string in mod is DamageForm now, maybe Unarmed should be form as well).
-                            if (Weapon.Types.ContainsKey(m.Groups[2].Value) && Nature.Is(Weapon.Types[m.Groups[2].Value]))
-                                moreAS += m.Groups[1].Value == "more" ? attr.Value[0] : -attr.Value[0];
-                            else if (DamageNature.Forms.ContainsKey(m.Groups[2].Value) && Nature.Is(DamageNature.Forms[m.Groups[2].Value]))
-                                moreAS += m.Groups[1].Value == "more" ? attr.Value[0] : -attr.Value[0];
-                        }
+                            if (Weapon.Types.ContainsKey(m.Groups[2].Value) && Nature.Is(Weapon.Types[m.Groups[2].Value]) || DamageNature.Forms.ContainsKey(m.Groups[2].Value) && Nature.Is(DamageNature.Forms[m.Groups[2].Value]))
+                                moreAS *= (m.Groups[1].Value == "more" ? attr.Value[0] : -attr.Value[0]) / 100 + 1;
```
l0g0sys: Same here. There is condition for RegExp subgroup "more" to treat mod as "more AS" otherise as "less AS"

PR #402, opened 2016-11-22
Summation of tree changes

Krumel:
I've added a summation of all the changes to attributes, to the tooltip that appears on hovering over a node, when clicking that node would change multiple nodes in the tree (adding or removing).

It's a bit clunky, but I think it would be a nice feature to have, as you can quickly compare clusters of nodes.

It also doesn't have any form of localization support atm, but so does the tooltip line directly above.

### DIFF
--- a/WPFSKillTree/Model/Options.cs
+++ b/WPFSKillTree/Model/Options.cs
@@ -12,6 +12,7 @@ namespace POESKillTree.Model
         private bool _characterSheetBarOpened;
         private bool _buildsBarOpened;
         private bool _treeComparisonEnabled;
+        private bool _changeSummaryEnabled;
         private bool _showAllAscendancyClasses = true;
         private string _nodeSearchHighlightColor = "Red";
         private string _nodeAttrHighlightColor = "LawnGreen";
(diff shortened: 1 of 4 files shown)


brather1ng:
I like the idea, I think it came up before already. However, I don't think it should always be on. It would need to be a toggle to not clutter the tooltip.

EmmittJ:
Yeah, this is something I had done before, but I never really finished the
idea. It definitely needs a menu option.

Krumel:
I've changed it to only show while pressing Ctrl.

EmmittJ:
So, I had some talk with @brather1ng, and we aren't sure ctrl is the best option (since ctrl + click is for zooming out). What about adding it to the menu as well as giving it a hotkey? This way you can toggle it on/off pretty fast without having to hold down a key.

Krumel:
Sure, I'll change it that way, it'll just take a bit of time for me to get into the code some more.
Is there any key you would prefer, or should I just pick one (or is there a system for rebindable keys)?

MLanghof:
...you bound it to what most programs use as "Search" hotkey?

Krumel:
It's an easily reached hotkey and not yet bound. 
And that's why I asked for they key I should bind it to,  I'm shit at choosing

SpaceOgre:
What about `Ctrl + g` as in Group? I agree that `Ctrl + f` is not the best. I think I will bind that to the search-box actually :)

brather1ng:
Seems to work fine 👍 

I think the menu entry would fit better under View, but that's just my opinion.

Krumel:
Okay, I've added both those suggestions in.
I've also restructured the code a bit to make githubs diff spazz out less and to make it easier to see what I changed.

EmmittJ:
Sweet! Looks good. I am going to merge this now!

PR #404, opened 2016-11-22
Add possibility to load tree as new build

brather1ng:
"Load Tree" button is a Split Button. Trees can either be loaded like they are now or as a new build.

Doing this as a PR to get opinions on the button functionality. I tried executing the commands when the button option is changed, but that let to 3 build creation popups per selection change and I think requiring the button to be clicked isn't much worse.

### DIFF
--- a/WPFSKillTree/Controls/CommandCollectionViewModel.cs
+++ b/WPFSKillTree/Controls/CommandCollectionViewModel.cs
@@ -1,0 +1,69 @@
+﻿using System;
+using System.Collections.ObjectModel;
+using System.Windows.Input;
+using POESKillTree.Common.ViewModels;
+using POESKillTree.Utils;
+
+namespace POESKillTree.Controls
+{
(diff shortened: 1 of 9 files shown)


EmmittJ:
I think the drop down should only change the function of the button (like it is currently). It works well.

SpaceOgre:
Nice feature!

Two things:
1. The choice should problably be saved in persistent data. I think most people will use one or the other most of the time depending on what they like, so it would be good to not have the "Load as new build" people have too re-select every restart.
2. The distance to the right edge of the drop down for "load as new build" is to small when Load Tree is selected: 
poe_skillplanner

brather1ng:
@SpaceOgre Done! Thanks for the input.

SpaceOgre:
Looks good to me, no more comments.

EmmittJ:
Ah, yeah. Good suggestions. Looks solid!

PR #407, opened 2016-11-26
Code improvements, added unit tests, ...

DrLeh:
I made several code improvements in order to be able to add more unit tests. The code was using a lot of classes as static that makes certain functionality impossible to test.

*  I split code for the `ItemDB` (now called `GemDB` into multiple code files for readability, since there wasn't really a reason to have everything be child classes. 
* I split the Compute class into several class files for readability. I renamed it `Computation` simply because I added a new folder and namespace titled `Compute` that was causing naming conflicts.
* I showcased how, by making these changes, the code is not more testable by adding some unit tests for resistances. 
* Miscellaneous refactorings and improvements for code reuse and usability.

From what I could tell, the app runs well as before, these changes were mostly for unit testing help. Please let me know if you have any feedback and I will improve this PR.

Thanks!


DrLeh:
Removed that constructor, reverted changes to that JSON file.

brather1ng:
Nice!

A few things related to the refactoring:

- The stuff from the previous `ItemDB` should get its own namespace (probably under `Model`). `SkillTreeFiles` is pretty filled now.
- The fields you had to add to other classes should follow our convention: `GemDB DB` in `GemLoader` should be `private readonly GemDB _db` and `Computation Compute` in `MainWindw` should be `private Computation _compute`. Same applies to all fields from the existing classes, if you feel like doing more refactoring, but those aren't your fault (there are also non-private fileds that can't be private and should be properties, everything else is not that important atm, imo).
- You added the NUnit testing framework for one test. Is there a reason for that? I don't think we should mix two testing frameworks.

I'll leave testing it until you are done, but no functionality should have changed anyway.

DrLeh:
thanks for reviewing it, i'll put those suggestions in place. I put in NUnit because i have NCrunch installed on my machine which makes running unit tests a million easier and faster.  I can remove it for now.

Side note: what version of VS do you use? Are you on 2015+ so we can use C# 6?

brather1ng:
> Side note: what version of VS do you use? Are you on 2015+ so we can use C# 6?

VS 2015, so yes

DrLeh:
@brather1ng  I've addressed your items. I moved the GemDB stuff to Models/Gems. I have a bunch of other things I want to do with unit tests, so I might end up adding NUnit back in replacement for the built in framework, but that can happen later.

EmmittJ:
I assume this is now considered a complete pr?

DrLeh:
Yes

EmmittJ:
Sorry I haven't gotten to this. I am currently dealing with finals. Maybe someone else could double check (@brather1ng) this. It looks good to me, but I haven't had time to pull this down.

brather1ng:
I'll check it out later today.

brather1ng:
* `TestCharacterSheet.Build_Test()` fails (most stats are wrong now)
* Program crashes on startup (`MainWindow.xaml.cs:446`) because that file does not exist (it should be named `ItemsLocal.xml` like it was before and it should not be loaded if it doesn't exist).
* With that fixed, it crashes when loading a build with items because the gem files are only loaded into a local variable not used anywhere else.
* Generic `Exception` should not be thrown (e.g. in those `IsIndexed` checks in `GemDB`, `UnsupportedOperationException` would be more fitting here)
* When you fixed those, you should actually run the program, load characters yourself and verify it's working like before.

PR #423, opened 2016-12-11
Searchable combo box

brather1ng:
Finally got a properly working searchable combo box (already tried it when I did the advanced generator). These combo boxes filter their drop down contents based on text typed there. Just click on the combo box and start typing.

Added to:

* Class filter in build list.
* Attribute and Pseudo Attribute selection in the advanced generator tab.
* Base item selection in item crafting window. Also added an "Any" base item group and type so the base item can be selected without selecting group and type first.

I want to add unique crafting next and I think unique selection would be pretty hard with a normal combo box, so this is kinda necessary.

Also, I reran the base item and gem loaders, which somewhat blows up the diff.

### DIFF
--- a/WPFSKillTree/Controls/SearchableComboBox.cs
+++ b/WPFSKillTree/Controls/SearchableComboBox.cs
@@ -134,15 +134,6 @@ namespace POESKillTree.Controls
 
         protected override void OnItemsSourceChanged(IEnumerable oldValue, IEnumerable newValue)
         {
-            if (newValue != null)
-            {
-                _itemsSourceView = CollectionViewSource.GetDefaultView(newValue);
-                if (_itemsSourceView.Filter != null)
-                {
-                    _oldFilter = _itemsSourceView.Filter;
-                }
-                _itemsSourceView.Filter = Filter;
-            }
             if (oldValue != null)
             {
                 if (_oldFilter != null)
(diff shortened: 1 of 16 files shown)


madhex:
I've tested it a bit and found the following issues:

1. In Craft Window when item Group is selected (first combo box), search stops working.

2. This one I got on all filters. If you open a combo box list and scroll it down, filter textbox can't get focus. Instead of that, it scrolls list up on each click as an up arrow does.

3. On both combo boxes in the Generator tab if I start typing then hover mouse over drop down contents and continue typing, combo boxes close themselves. It occurs because of focus loss too.

P.S. As this is my first appearance on the repository, I’d like to say thank you to all participants for sharing and supporting this project and wish good luck at further development. Cheers!

brather1ng:
Thanks for checking it out and for the nice words!

1. Fixed
2. I couldn't find a way to fix it. I don't even know what causes this behaviour, at least none of the mouse event handlers the combo box has access to.
3. Can't reproduce this. ("click box -> type -> hover an item -> type" works as expected)

brather1ng:
Someone else wants to check this out? If not, I'll merge it in a few days, it's nothing major anyway.

EmmittJ:
Hey, I can get to it in a couple days. Holiday season is a bit difficult to
get to stuff.

brather1ng:
No worries, I don't think it necessarily needs to be checked out, it's nothing big. Was only asking if someone already planned to look into it. So no problem if you don't find the time.

EmmittJ:
So, this looks awesome. They work as expected, or at least as far as I can
tell.

On Dec 27, 2016 6:41 AM, "brather1ng" <notifications@github.com> wrote:

No worries, I don't think it necessarily needs to be checked out, it's
nothing big. Was only asking if someone already planned to look into it. So
no problem if you don't find the time.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/EmmittJ/PoESkillTree/pull/423#issuecomment-269334333>,
or mute the thread

brather1ng:
Thanks!

PR #433, opened 2017-01-26
PoE Planner urls support (Fixes #427) and other ulr decoding enhancement

### LINKED ISSUE #427: poeplanner url gives "index outside bounds of array"
http://poeplanner.com/AAQAAPcTEAB534rpAkuuGyXZW20ZEVA1uacrHU-X9FJTPV-XlUyz0B_v6yBuUDARD6yY99eESFxrw23i9wFvi3oi9GwLkyd_xsHF7Bgsv4_60NBfakcG2-e0xcM6KjhcQFS9Fe2Wi9W5h8unm3WezxXYdrk-rGYs4X3j8uFjQ2wI8YrviPk3189Jsb6KoS8RlpAbcFaaE9vUieDmWJMfSRPK07QM6roNfBXXhq5WYw5IXfIqC3BSoqMPxDt8uMrw1UlROw32_Kc0YqxsjJ2qES_rY7VI1EKEb3X9Fr8PqySLNj1W9dtul9COihpI9Ch88JEHu_wVfn6hadgAAAAAAA==

Thanks for a great program, I use it all the time!

madhex:
### Features

- Loading from poeplanner (fixes #427), including bandits selection
- Goo.gl URL shortener suppport
- Improved auto-completion of supported URLs, so it is now possible to use URLs without protocols like poeurl.com/ZfJ

### Technical details

The idea was not only to revive poplenner support but also reduce SkillTree responsibility and make it less godlike.

#### Urls processing

I've moved existing logic of Urls decoding to separate classes called deserializers, and added BuildConverter which is static factory that automates their selection. BuildUrlDeserializer is the base abstract class, it wraps build URL and provides particular Ids, class names and aggregated object (BuildUrlData), representing raw decoded information. Derived classes:
 - PoeplannerUrlDeserializer - extracts supported data and also does some pre-processing for jewels and minimal for equip. I believe it may be useful someday.
 - NaivePoEUrlDeserializer - default one which is supposed to be used as a fallback when the others deserializers refused to parse specified Url. There are many PoE builders on the internet that copies the official site and their Urls are completely compatible. This deserializer actually encapsulates logic for official site, but has very weak Url check, so it can at least try to process with unpredictable outcome. It is used as the last one.
 - PathofexileUrlDeserializer - derives from NaivePoEUrlDeserializer. Validates Url, skill tree version and promotes processing to naive if possible.

The ISkillTree interface was declaring methods that accepts Urls and doesn't require skill tree state, so they were moved to deserializers too.

Serialization logic was moved from SkillTree.SaveToUrl to the new class SkillTreeSerializer. It is a wrapper that consumes an instance of the SkillTree or BuildUrlData class and converts it to a pathofexile.com Url. The serializer also provides methods for generating Urls of empty starting builds.

SkillTreeImporter used to fetch poezone.ru builds is removed as obsolete. It appeared that their skill tree is inactive since 2012, and changes in repository were driven by renamings and refactorings.

#### Bandits support

The PoESkillTree and PoE Planner uses different identifiers for bandits, so I have explicitly specified numeric values in Bandit enum to state that they are essential and added BanditConverter that holds casts. This class is used by deserializers.

When PoE Planer Url is loaded and it contains bandits different from currently selected, User will be asked whether to overwrite existing selections ot not.

#### Tests

For the purpose of testing I've created BuildUrls.xml and EmptyBuildUrls.xml files and corresponding classes for their loading. Not sure if used approach is quite suitable for the project, but in case it is, I've also created xsd file to validate these xml and help in their maintenance.

I modified the existing build from the Builds.xml as it appeared to be obsolete.

#### Performance notes

As new deserializers are using more specifiс and precise regular expressions for validation and parsing, they appear to be much faster than the existnig. I've created fake build containing all 1368 nodes +30 Scion ascendancy nodes, on my local machine it required ~5.7 seconds to load previously vs ~0.7s with the new approach. However, for normal trees results are more modest: 0.22s vs 0.16s. So I had to mention that, otherwise you would not know about it =)


WPFSKillTree/SkillTreeFiles/SkillTree.cs
```diff
@@ -1171,62 +1158,60 @@ public void CrossAllHighlightedNodes()
...
             }
 
-            foreach (var node in nodes)
+            foreach (var nodeId in buildData.SkilledNodesIds)
             {
-                skillednodes.Add(Skillnodes[node]);
+                SkillNode node;
+                if (!Skillnodes.TryGetValue(nodeId, out node))
+                {
+                    throw new KeyNotFoundException(L10n.Message("The build you are trying to load contains unknown nodes."));
```
EmmittJ: This will happen every league, and we will have a very unhappy user base. They generally like to convert, or reference, old builds, and this has been a feature for sometime now.

WPFSKillTree/Utils/UrlProcessing/NaivePoeUrlDeserializer.cs
```diff
@@ -0,0 +1,92 @@
...
+        public override BuildUrlData GetBuildData()
+        {
+            var bytes = GetRawData();
+
+            var deserializedData = new BuildUrlData(BanditConverter.Default);
+            deserializedData.Version = bytes[0] << 24 | bytes[1] << 16 | bytes[2] << 8 | bytes[3];
+
+            if (!IsVersionSupported(deserializedData.Version))
+            {
+                throw new NotSupportedException(
```
EmmittJ: Again, we normally don't do this type of thing because people like to looks at builds between leagues as well as load older versions of builds.

WPFSKillTree/SkillTreeFiles/AscendancyClasses.cs
```diff
@@ -3,14 +3,9 @@
 
 namespace POESKillTree.SkillTreeFiles
 {
-    public class AscendancyClasses
+    public static class AscendancyClasses
```
brather1ng: I'm not really a fan of static state, makes testing harder because of hidden dependencies.
Rather make those available through `ISkillTree` if necessary (that would be great, as it would lead to lower dependence of the build view model stuff on `SkillTree` than before your work).
Yes, SkillTree has a lot of static state. But that should be reworked at some point and it actually partly causes the lack of tests (e.g. I couldn't write tests for most of the tree generator stuff because of it).

(same thing with `BuildConverter`)

WPFSKillTree/Utils/UrlProcessing/BuildUrlNormalizer.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+            else
+            {
+                response = await new HttpClient().GetAsync(buildUrl, HttpCompletionOption.ResponseHeadersRead);
+            }
+
+            return response.RequestMessage.RequestUri.ToString();
+        }
+
+        protected virtual string EnsureProtocol(string buildUrl)
```
brather1ng: There are a bunch of methods that are only visible because they are used for testing (not only in this class). I'm kinda conflicted on that, optimally all functionality could be tested over the real public API, but it's better than no tests so it wouldn't stop me from accepting this.

WPFSKillTree/SkillTreeFiles/SkillTree.cs
madhex: This sounds reasonable. I'll remove exception throw. But I find it a bit inconvenient for inexperienced users and players to skip build issues silently. I could extend the BuildUrlModel with some field (IsValid, more informative State, etc.) and show notification or warning in MainWindow if build is not Ok. What do you think?

BTW, i'm quite new to the game, so correct me if I'm wrong, but as far as I see tree and particular nodes ids doesn't change between each league and it is the same tree introduced in 2.1.0. So it does change, and we have to deal with it, but no so often.

WPFSKillTree/Model/Bandit.cs
```diff
@@ -8,14 +8,14 @@ namespace POESKillTree.Model
...
     public enum Bandit
     {
-        None,
-        Alira,
-        Oak,
-        Kraityn
+        None = 0,
+        Alira = 1,
+        Oak = 2,
+        Kraityn = 3
```
madhex: I've tried it during implementation and it worked. I've searched all occurrences manually, and they looked fine. I've wrote some odd tests and they passed. But I'm still not sure that I can oversee everything, e.g. there might be some old *.pbuild files that were deserialized using numeric values or other xml/json (code that created compatible content but now burried in repo history) or something else, that I'm not aware of.
Moreover, some day we can get another great online planner with it's own format, or update for the official planner which values would be more preferable for us to use I believe.
Keeping all that in mind I've left values as is, but if you think this is overhead I'll remove whole `BanditConverter` and will just switch values.

WPFSKillTree/SkillTreeFiles/SkillTree.cs
```diff
@@ -1171,62 +1158,60 @@ public void CrossAllHighlightedNodes()
...
-            var nodes = new List<ushort>();
-            for (int k = (i > 3 ? 7 : 6); k < decbuff.Length; k += 2)
-            {
-                byte[] dbff = { decbuff[k + 1], decbuff[k + 0] };
-                if (Skillnodes.Keys.Contains(BitConverter.ToUInt16(dbff, 0)))
-                    nodes.Add((BitConverter.ToUInt16(dbff, 0)));
-            }
-            chartype = b;
-            asctype = asc;
+        protected static BuildUrlData DecodeUrlProtected(string url, out HashSet<SkillNode> skillednodes, out int chartype, out int ascType)
```
madhex: No reasons. I'll rename it to `private ... DecodeUrlPrivate`.

WPFSKillTree/Utils/UrlProcessing/BanditConverter.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using POESKillTree.Model;
+
+namespace POESKillTree.Utils.UrlProcessing
+{
+    /// <summary>
+    /// Provides functionality for converting bandits identifiers from external sources into used in the POESKillTree application.
+    /// </summary>
+    public class BanditConverter
```
madhex: This delegate makes code reusable, but the class in question prevents copy-paste better and adds some degree of compile-time protection.
Since C# compiler is capable of inferring delegate type it makes possible to pass any suitable method or lambda. In this case if someone needs an instance of the BuildUrlData class and is not aware of BanditConverters, it is possible to replicate the same switch/case logic due to ignorance or even worth to write something erroneous like `new BuildUrlData(id => (Bandit)42)` that would fail only in run-time. On the other hand it is much harder to ignore required class.
Therefore, if this class survives other reviews, I would strongly recommend to accept it as is.

WPFSKillTree/Utils/UrlProcessing/BuildUrlNormalizer.cs
madhex: I've made a lot of methods `protected virtual` thinking first of all about extensibility and then of testability. Therefore, I've made tests for these methods just because it was possible and not vice versa.

WPFSKillTree/SkillTreeFiles/SkillTree.cs
brather1ng: I think a warning is ok if it only appears once per build.

The tree didn't change since Ascendancy, I think. But before that, major tree changes were pretty common.

WPFSKillTree/SkillTreeFiles/AscendancyClasses.cs
madhex: The hardest one...
Ideally I also would like to avoid new static classes introduction. To reduce problems while usage and testing I made them re-initializable. `BuildConverter` (assuming you advice for default deserializer) doesn't hide its dependencies: every delegate it calls it receives from the outside. `AscendancyClasses` looks worse.
I'll see what I can do but I assume it will be just different at the moment, not better. The SkillTree is responsible for initialization of lots of things and interfaces extraction doesn't realy help, because you still need to access the SkillTree to get needed reference while these static classes available independently.

WPFSKillTree/Utils/UrlProcessing/BuildUrlNormalizer.cs
brather1ng: Fine for me then.

WPFSKillTree/Utils/UrlProcessing/BanditConverter.cs
brather1ng: Hm, it would look lot cleaner imo and I don't think someone would misinterpret it. But I wouldn't really mind if it stay as is.

WPFSKillTree/SkillTreeFiles/AscendancyClasses.cs
brather1ng: It's more about dependencies from other classes on these. If they are passed as instances (e.g. via `ISkillTree`), they can just be mocked in tests. When they are used statically, there's no way but to deal with them.

If you can't change it easily, you don't have to. It's not worse than before, it would just be a good way to make it better, imo.

WPFSKillTree/SkillTreeFiles/SkillTree.cs
EmmittJ: Generally the node ids don't change much inbetween (unless there is a major rework). 
There isn't much we actually have to deal with since we use a hash it is O(1) to check if a node is there, and if it isn't move on.

WPFSKillTree/Utils/UrlProcessing/BuildConverter.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        private static ISet<TryCreateDeserializer> _deserializersFactories = new HashSet<TryCreateDeserializer>();
+        private static Type _defaultDeserializerType;
+
+        /// <summary>
+        /// Registers default deserializer that should be used if all others refused to parse.
+        /// It may be useful when trying to parse obsolete build urls and official site compatible urls.
+        /// </summary>
+        /// <param name="defaultDeserializer">The default deserializer type.</param>
+        public static void RegisterDefaultDeserializer(Type defaultDeserializer)
```
madhex: It looks nice!
I'll use it with constraint like this:
`public static void RegisterDefaultDeserializer<T>(Func<string, T> factory) where T : BuildUrlDeserializer`

madhex:
I've updated errors handling, so the application shows warning if loaded build has some unknown nodes or uses old Url format. This warning appeares only once and then Url is converted to a relevant version.
`BuildConverter ` and `AscendancyClasses` were refactored to non-static classes with appropriate interfaces. I made them available as properties of the `ISkillTree`.

Testing experience deteriorated because now there are more cases for which a static skill tree state should be initialized. Some more changes are requred to handle that, e.g. `SkillTree.RootNodeList` and `SkillTree.Skillnodes` are still static, but these properties are used too widely and I fill that this is too far from the scope of my original PR.

WPFSKillTree/ViewModels/BuildsControlViewModel.cs
```diff
@@ -196,7 +196,7 @@ public ISkillTree SkillTree
         public IReadOnlyList<ClassFilterItem> ClassFilterItems { get; }
 
         public BuildsControlViewModel(IExtendedDialogCoordinator dialogCoordinator, IPersistentData persistentData,
-            ISkillTree skillTree)
+            ISkillTree skillTree, IAscendancyClasses ascendancyClasses)
```
brather1ng: No need to pass `ascendancyClasses`, the are right there in `skillTree`.

EmmittJ:
Not anywhere where I can test this, but what happens if you load a shortener (i.e., poeurl) with that shortened a poeplanner url?

brather1ng:
Looks good to me now (except that one really minor thing I commented). I'll play around with it a bit.

I don't think the tests got worse. If I'm not mistaken, the tests you changed were dependent on skill tree initialization before (to initialize `AscendancyClasses` and `BuildConverters`). It only got a bit more obvious and safer now. And yes, `SkillTree` in general is pretty bad, but yeah, that's pretty out of scope.

WPFSKillTree/Utils/UrlProcessing/BuildUrlDeserializer.cs
```diff
@@ -0,0 +1,88 @@
...
+using POESKillTree.SkillTreeFiles;
+
+namespace POESKillTree.Utils.UrlProcessing
+{
+    /// <summary>
+    /// Represents an object that extracts build information from the tree url.
+    /// </summary>
+    public abstract class BuildUrlDeserializer
+    {
+        protected static IAscendancyClasses AscendancyClasses;
```
brather1ng: I think you forgot to remove the `static` here.

brather1ng:
@EmmittJ works fine. Shortener resolving happens before the urls are converted to trees/official urls.

WPFSKillTree/ViewModels/BuildsControlViewModel.cs
madhex: Right. It looks like I've left some inconsistency while playing with different interface dependencies.

madhex:
@EmmittJ build loading process is as follows:
1. Normalizer tries to make Url acceptable to deserializers: it trims provided Url, decodes encoded Urls (described bellow), extracts google query parameter if needed, unfolds shorteners (including shortened shorteners), prepends protocols.
2. One of deserializers extracts data from that normalized Url (official or poeplanner).
This step was made to load additional information provided in unofficial formats.
3. Url is converted to the official format and assigned to build which leads to notification event and normal nodes loading.

This is similar to the existing process where Url is corrected iteratively until application is able to load it. Therefore, shortened poeplanner Url works.

#### Update
It turnd out that the official site has a bug that breaks some Urls when swithing to localized pages. Some symbols appeared to be encoded:
- expected (en) build ending
...H2_PfX-Tc<b>=</b>
- actual ru/br builds ending
...H2_PfX-Tc<b>%3D</b>

To fix this and similar cases I've added additional encoding step to the normalizer.

EmmittJ:
Good stuff! Seriously, thank you!

EmmittJ:
@brather1ng any objections to merging. I see you have approved this

brather1ng:
Looks good to me, go ahead.

And a big thanks to @madhex!

PR #435, opened 2017-01-31
Use Wiki API or GGPK instead of Wiki HTML parsing

brather1ng:
I have rewritten all UpdateDB loaders to use either the Wiki's API or the GGPK data. I first wanted to do everything with the API, but for gems that is not entirely possible, so I needed the raw data.

This PR itself doesn't provide any new functionality. It only makes maintaining the extraction code easier, and I wanted to have a basis that is known to be correct to use for extracting uniques. This also means that merging this isn't really important atm.

For the GGPK data, I have created a new repository ([RePoE](https://github.com/brather1ng/RePoE)) where the files are extracted with PyPoE and converted to JSON, but that's another thing.

- Rewrite ItemDataLoader and ItemImageLoader to use the Wiki's API instead of parsing their pages
- Rewrite GemLoader to use the GGPK data. The raw data has enough information that most of the attributes that were formerly in `GemProperties.xml` can now be extracted automatically (e.g. weapons restrictions and dual wield behaviour).
- Adjust the in-program item image loading to use the Wiki's API instead of parsing their pages.
- Add services for the in-program item image loading to guarantee each image is only loaded/downloaded once.


UpdateDB/DataLoading/DataLoader.cs
```diff
@@ -18,6 +18,12 @@ public interface IDataLoader
         bool SavePathIsFolder { get; }
 
         /// <summary>
+        /// Sets the <see cref="HttpContent"/> used for downloading data.
```
MLanghof: Did you mean HttpClient?
I see that it was previously called that in other places, but if it's intended I can't make sense of it.

UpdateDB/DataLoading/ItemDataLoader.cs
```diff
@@ -1,255 +1,389 @@
...
+using POESKillTree.Utils.WikiApi;
+
+using static POESKillTree.Utils.WikiApi.WikiApiUtils;
+using static POESKillTree.Utils.WikiApi.ItemRdfPredicates;
 
 namespace UpdateDB.DataLoading
 {
     /// <summary>
-    /// Extracts item bases from the wiki as a <see cref="XmlItemList"/>.
+    /// Retrieves item bases from the unofficial Wiki through its API.
```
MLanghof: Technically it's linked to from pathofexile.com (see the "Game" dropdown in the menu bar), and it consequently calls itself the "Official Path of Exile Wiki".
This also appears in a few other places.

WPFSKillTree/Model/Items/EquipmentData.cs
```diff
@@ -11,19 +11,20 @@ namespace POESKillTree.Model.Items
...
     {
-        private readonly Options _options;
 
         public IReadOnlyDictionary<ItemType, IReadOnlyList<Affix>> AffixesPerItemType { get; private set; }
 
         public IReadOnlyList<ItemBase> BaseList { get; private set; }
 
         public IReadOnlyDictionary<string, ItemBase> BaseDictionary { get; private set; }
 
+        public ItemImageService ItemImageService { get; }
```
MLanghof: I didn't even know this was allowed by the language... Maybe an actual accessor function + private variable would be less... confusing here? (I totally get how it's consequential in its current form)

WPFSKillTree/Utils/Extensions/NumberExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+        public static bool AlmostEquals(this float x, float y, float epsilon)
+        {
+            return Math.Abs(x - y) <= epsilon;
+        }
+
+        public static bool AlmostEquals(this double x, double y)
+        {
+            // double has 15 significant digits. 13 leaves a bit of room for accumulated errors.
+            var epsilon = Math.Max(Math.Abs(x), Math.Abs(y)) * 1e-14;
+            return AlmostEquals(x, y, epsilon);
```
MLanghof: This (and also the float one above) won't work properly if one value is e.g. zero and the other is a denorm (epsilon will be zero and they will therefore be considered unequal).

You're essentially trying to figure out how many ULPs the smaller number is away from the larger one. But ULP-based comparison (as demonstrated) breaks down near zero. [This article](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/) demonstrates it quite well. Since you're actually going through the trouble of writing this extension, I'd probably recommend [this](http://floating-point-gui.de/errors/comparison/) approach.

MLanghof:
On an unrelated note, is there a cure for the godawful way in which git handles { and } in diffs? You can rewrite an entire file, but the diff will just "randomly" pick out opening and closing curly braces to latch onto and then garble the previous/current comparison because of it.

UpdateDB/DataLoading/DataLoader.cs
brather1ng: Yes, no idea were the `HttpContent`s came from.

WPFSKillTree/Utils/Extensions/NumberExtensions.cs
brather1ng: Thanks, didn't know that.

These were used exactly once, so I'll keep it simple and just remove the relative error ones. In most cases, it should be easy to manually set a delta that makes sense in the context (e.g. stats only use up to 2-3 decimal places).

WPFSKillTree/Model/Items/EquipmentData.cs
brather1ng: Hm, I think this is used quite commonly in C#. It's much less confusing in an IDE than in the diff because class and instance are colored differently.

EmmittJ:
@brather1ng looks like the conflicts should be pretty easy to resolve here

brather1ng:
@EmmittJ conflicts resolved

brather1ng:
@EmmittJ do you want to look over this or can I merge it?

EmmittJ:
Sorry, I haven't had time. If you feel it's ready go ahead otherwise I can
look at it later today.

brather1ng:
No problem, @MLanghof looked over it, so it's fine.

EmmittJ:
Yeah, I saw that. Thanks :)

MLanghof:
To note, I didn't actually test it (since I'm entirely green on the topic/process), just looked over the code.

Issue #426, opened 2016-12-21
Pseudo Attributes Empty - what to download?


daAresinger:
So the planner keeps crashing when I start the programm.

The stacktrace tells me to check the PseudoAttributes folder. It turns out to be empty, which it apparently shouldn't be. What do I need to download to be able to fix this?

[Sorry didn't feel like creating a Pastebin acc for this.](https://drive.google.com/open?id=0By037qa7xxEwNXhrSkY5c0dhSzA)

brather1ng:
This appears frequently, so I wrote a wiki page: [Program crash: "No Pseudo Attributes loaded."](https://github.com/EmmittJ/PoESkillTree/wiki/Program-crash:-%22No-Pseudo-Attributes-loaded.%22)

Is that the only empty data directory (do `Data/Equipment` and `Data/ItemDb` contain XML files)?

daAresinger:
yes those two are empty aswell.
Thanks for the help

PR #436, opened 2017-02-18
Add unique item crafting

### LINKED ISSUE #157: Crafting windows - Uniques
Unique's  should be selectable in the crafting window

brather1ng:
Uniques can be crafted over the new button on the equipment tab. It works similar to the normal/magic/rare crafting.

It is also possible to craft drop disabled items (both uniques and other). Drop disabled items are hidden by default. I'm not sure about the placement of the toggle for that.

Uniques are retrieved from the wiki, similar to item bases. Their images are downloaded in addition to those of item bases. This increases the image folder's size from about 10 to about 16 MB, which is still completely fine imo.

I have refactored the crafting related controls to MVVM.

Some other various improvements:
- only bases where we actually have affixes for are shown in the crafting windows
- only groups and types with at least one base are shown in the crafting windows
- inventory height and width of items is no longer hardcoded but retrieved from the wiki
- All equipment lists are now resources. This means they will not be missed by a misbehaving installer (i.e. as in #426).
- "Adds # to # damage" Stats are no longer split into minimum and maximum as supporting stats with multiple values is necessary for uniques anyway.
- Improved stat value sliders:
  * Their text is bound to their values so the text dynamically updates.
  * They are grouped if one stat has multiple sliders. The group has the stat text and no slider has text in that case.
  * Sliders are hidden if the value is fixed (the text is still shown).
- Stats are kept when changing to an item with the same possible affixes (i.e. when changing to an item base of the same type, does not affect uniques)

Known issues:
- Some stats have multiple lines. These are retrieved from the wiki as separate stats. They could only be combined if the actual mod ids were retrieved and translated by us, which would be a lot more work. This isn't really a problem, as these long stats are not supported by our calculation stuff anyway.

Closes #157


UpdateDB/DataLoading/AffixDataLoader.cs
```diff
@@ -233,13 +235,42 @@ private static IEnumerable<XmlStat> ExtractStats(string statColumn, string affix
...
+                    From = new[] { from },
+                    To = new[] { to }
+                });
+            }
+
+            // merge "... # minimum ..., ... # maximum" into "... # to # ..."
+            if (xmlStats.Count > 1)
+            {
+                var previous = xmlStats[0];
+                for (var i = 1; i < xmlStats.Count; i++)
```
MLanghof: So I just looked this up and apparently ReSharper does this kind of thing. I certainly don't want to tell you how to write your code, but I'd just like to point out that I found this method (the first major one I'm looking at) very hard to grasp, mainly because of the `var` spam. I don't mean that I don't understand a for loop (or that the code overall is hard to understand), but it's kind of like taking the colour out of a photo - of course you can still see what's going on, but it's certainly less enjoyable for others to look at.

Again, I'm not the one writing 10k LOC (and I realize half this function was already like this). I don't want to suggest you turn off that feature, just leaving the feedback here that it shifts the burden of "readable code" from the writer to the reader (in my eyes). Maybe the variable names could be more descriptive, but maybe that would distract more from what's going on. I don't know.

MLanghof:
I'll see if I can find the time to test it. Good to see the progress on a full-on "character builder" though!

UpdateDB/DataLoading/AffixDataLoader.cs
brather1ng: Yeah, ReSharper's default settings suggest using var for everything. It's only marked as "Hint" though (not as a code smell).

It's probably just me being lazy in most cases. I should definitely change my habit of using var everywhere and use only use it where the type is obvious. Changing int to var in loops doesn't really make sense at all, it doesn't even save space.

MLanghof:
Alright, took a look.

- Is the text supposed to be [on top of the sliders](https://i.imgur.com/nMx9U7z.png)? Seems pretty hard to read for me.

- The item display [doesn't use enough of the horizontal space](https://i.imgur.com/DZ6R1eX.png) for me.

- Item base (and every other) setting is not persistent. Not sure if that's relevant for normal use, just mentioning it though.

- I'd replace "Crafted by PoESkillTree" by something like "Created with PoESkillTree" to avoid confusion.

- I'm not sure if things like total phys damage values are rounded down or to nearest int.

- Ventor's gamble (and similar items) isn't displaying its affixes properly with respect to "reduced"/"increased" and the +/- for resistances (and, I guess, hiding affixes with value 0?). Don't know how much of this is fixable (are the other affix "displays" available anywhere?).

- The unique list tends to be quite sluggish if no base type is selected.

Other than that it's nice and sweet!

PS: Thanks for going over the var stuff btw, so much better now.

brather1ng:
> Is the text supposed to be on top of the sliders? Seems pretty hard to read for me. I feel like it's not supposed to (since it's different for dual-value stats), but I can't really tell.

Was intentional, as that's what it was like before. But it is easier to read when separated, so I changed that.

> The item display doesn't use enough of the horizontal space for me.

This is a hard one. If the main crafting part in the middle should be centered (which looks best), the left side with the display has to be of the same size as the empty right side. I have limited the size of the middle part, the window has to be kinda small now for the display not to fit.

> Item base (and every other) setting is not persistent between dialogue openings. Not sure if that's relevant for normal use, just mentioning it though.

Yeah, the window is recreated each time. I don't think anyone would want to craft multiple similar items?

> Ventor's gamble (and similar items) isn't displaying its affixes properly with respect to "reduced"/"increased" and the +/- for resistances (and, I guess, hiding affixes with value 0?). Don't know how much of this is fixable (are the other affix strings available programmatically anywhere?).

The +/- for resistances was a bug.

The information to switch between "reduced"/"increased" is not available, I would need to use the game's mod and stat ids and translate them to text representations myself. I might do that at some point, but it's too much work for just a few uniques, imo.

On hiding affixes with value 0: I'm not entirely sure if that's always the case and I don't think it's worth going through all affixes to check it. When stat ids are used, this would be possible (hide the stat if it has no translation matching the values).

> The unique list tends to be quite sluggish if no base type is selected. Unsure if this can be improved.

The built-in way to filter wpf CollectionViews (its Filter method) leads to all controls in the list being rebuild on each filter change. This is slow for multiple hundred items. I use that for the searching, so if you have no base type selected, the filter is reapplied on all items, which is slow. I could probably change it to hide filtered combo box items instead of using the built-in filter, which rebuilds the combo box's popup without the filtered items, but I don't think that's easy.

> Weapon range is in a different position than on the wiki. ¯\_(ツ)_/¯

Is it? The elemental/chaos damage properties are added at the end, maybe that's what you saw. But yeah, I don't think that's worth fixing.

EmmittJ:
My apologies. I haven't had a chance to pull this down, and there are
enough changes that I can't just browser the code. I will try and get to
this tonight.

EmmittJ:
Honestly, this is working great. The fixes that @MLanghof suggested definitely help. The separated min-max sliders are good as well. I didn't have any problems, and I see jewels are loaded up which is a great start to implementing them. 

* I am not too sure if we need the hide drop disabled items toggle though since all items can be obtained in some way at this point with zana (as far as I know).

MLanghof:
> all items can be obtained in some way at this point with zana (as far as I know).

"Drop disabled" also applies to race trophies, datamined uniques etc. as far as I'm aware.

> If the main crafting part in the middle should be centered (which looks best), the left side with the display has to be of the same size as the empty right side.

Use the golden ratio (while occupying the whole space)? That's bound to look good! :P

> The elemental/chaos damage properties are added at the end

Yep, that's what I meant.

> I don't think anyone would want to craft multiple similar items?

I could imagine someone noticing a mistake on the item he made and wanting to redo it. Again though, this is pretty much irrelevant, I agree.

brather1ng:
"Drop disabled" other than league uniques should only be race rewards. Uniques can't be datamined. It's more useful for normal items: talismans bloat the amulet list and there are the old quiver bases. I don't know if the toggle should be there only for that.

> Use the golden ratio (while occupying the whole space)?

That looks worse with larger window sizes imo and only provides a bit more space with smaller ones.

> I could imagine someone noticing a mistake on the item he made and wanting to redo it. 

For that I'd like to at editing functionality at some point.

brather1ng:
I think I'll keep the drop disabled toggle. The drop disabled items are all special enough that they will not be crafted often.

Drop disabled uniques are race rewards, Star of Wraeclast, Descent Champions starter uniques, quiver uniques that weren't kept for the new bases, and Awakening beta only jewels. Drop disabled item bases are the bases of drop disabled uniques that can't be acquired normally, old quivers and Keyblade.

However, I'm still not sure about the placement of the toggle. It is quite prominent for the little relevance it has.  

I'll merge this in a day or two if there are no more remarks. Maybe I get an idea for a better toggle placement until then.

EmmittJ:
It could be placed off to the right hand side as a check box or something.
If there were more settings, it would definitely look better.

brather1ng:
I think I found a good spot:

dropdisabledplacement

EmmittJ:
Very nice. It cleans up a lot of space as well

PR #445, opened 2017-03-02
Added zh-TW Locale

Chuanhsing:


brather1ng:
Thanks for the translation!

There are at least our installer strings that should also be translated, see `Languages/` and `Default.isl` in  `WPFSKillTree\Tools\isetup\` (you can find presets at http://www.jrsoftware.org/files/istrans/), and you can add the translation to `WPFSKillTree/release.iss:Languages`. I'm not sure if that's all that is required to make translations work, though.

Chuanhsing:
I had tested this file and worked when putted it under Locale/. The program will automate select zh-TW Messages.po file.

brather1ng:
Yes, with the portable version. But most people install the program using the installer, which also has messages that can be translated (those are only a few additional lines if you use a preset for the common InnoSetup messages).

PR #447, opened 2017-03-03
Extended german translation

Chelyocarpus:
Extended and corrected german Messages.po. Still not perfect but i think it's an improvement. 
I've noticed that some terms and sentences from the interface are not in Messages.po so it's like a nice mix of english and german.

I hope it's useful.


brather1ng:
Looks good!

If you remember particular terms that are not in Messages.po, please report them to us. If they don't come directly from game data, they are bugs.

PR #451, opened 2017-03-16
Gem sockets

### LINKED ISSUE #335: Request - Place to add Gem Links
I really love this program but there just seems to be this one thing missing to make it perfect for creating builds I would like to play.

I don't think it needs to calculate into the stats, but would it be possible to add a section to place gem links. 

Maybe 2 lines of 6 gem slots and 4 line of 4 gem slots. (I think that would cover it all)

I just hate having to swap out to notepad or something else to look up what gems I want to run.

Either way, thanks for your hard work!

brather1ng:
Implementation of gem sockets as discussed in #335. You can get there via context menu on items in the inventory and stash.

I've also rewritten the inventory and stash controls to MVVM as I had no idea how they, the stash in particular, worked and disabling context menu entries (e.g. when the item has no socketes) is much easier this way. This is the majority of the diff, but it shouldn't affect any functionality.

Closes #335


MLanghof:
I promise to take a look!

MLanghof:
(Still really want to, but will take me at least until the weekend to find time for it - sorry!)

EmmittJ:
I am in the same boat (finals week)

MLanghof:
Alright, I found some time. I wasn't aware how much of this functionality for gems was already implemented. It looks like 80% of the diff is code cleanup and refactoring (which is great!). The new files also have very descriptive function names, I found everything easy to figure out.

In terms of testing it, here are things I noticed:

* Most importantly, you can resize the header cells, this shouldn't be possible: 

* For gem levels with multiple digits a scroll bar sneaks into the cell. There are also two bevels around the table's headers (or something like that). 

* The socket group can't be edited after adding a gem, you need to remove and readd it. Not sure if there are technical reasons against this.

* The vertical scroll bar that becomes necessary for 6 gems for me also makes the horizontal one appear:   
(Relatedly, clicking the x to remove a gem that is not fully within the view and also isn't selected will select and scroll to that item and (since the cursor is no longer on the x) fail to remove it).

* Also in that image, it seems that even group numbers are slightly grayed out while odd ones are not. That's not intuitive (and I don't know if there's a purpose for it).

* You can input floating point values for level and quality (and group, but that one gets validated away).

* I'd propose adding double-click to edit gems. I kept intuitively wanting to do it, and the right click -> menu item selection is mildly awkward (at least compared to the ingame drag-and-drop intuitiveness).

* If at all possible/reasonable (also considering the performance of the gem list), gem icons would be very awesome/helpful. But I realizable that this would also be unlike the rest of the equipment interface that does not bother with anything but the shapes.

brather1ng:
Thanks for the testing!

> For gem levels with multiple digits a scroll bar sneaks into the cell.

Doesn't happen for me. Gave them 5 pixel more width, did that help?

> There are also two bevels around the table's headers (or something like that). 

There's always an empty column at the end, I just made it small enough to almost disappear. Because of disabling resizing, these vertical stripes are gone anyway.

> The socket group can't be edited after adding a gem, you need to remove and readd it. Not sure if there are technical reasons against this.

There was a technical reason: gems in the collection were sorted by group, so I'd had to resort the collection on changes. I changed it to use a CollectionViewSource that is sorted by group. They are now sorted while typing, not sure if that's better (the code doesn't get any notification when the user moves focus out of the edited box, so it has to be resorted instantly on changing the value)

> The vertical scroll bar that becomes necessary for 6 gems for me also makes the horizontal one appear: 

That doesn't happen for me. I gave the DataGrid more space and set the horizontal scrollbar to always be hidden. Did that fix it?

> (Relatedly, clicking the x to remove a gem that is not fully within the view and also isn't selected will select and scroll to that item and (since the cursor is no longer on the x) fail to remove it)

I have no idea how to fix that. Better to make sure there is always enough vertical space.

> Also in that image, it seems that even group numbers are slightly grayed out while odd ones are not. That's not intuitive (and I don't know if there's a purpose for it).

That was a really bad try at making the groups easier to distinguish. It's hard to find fitting colors because accent and theme can be changed. The only dynamic colors available are the theme colors which look greyed out, and the accent colors, which clash with the hovered/selected backgrounds.

> You can input floating point values for level and quality when editing them

That is an issue with all up-down controls currently in the program. The floating point values are only visual, they are rounded. Didn't find an easy way to fix this.

> If at all possible/reasonable (also considering the performance of the gem list), gem icons would be very awesome/helpful.

That is implemented. The images are downloaded in the release script, so you don't normally get them. However, there is an option in the settings to download them: "Equipment -> Automatically download missing item images". The images for all base items, uniques and gems will be downloaded after restarting (might take a bit).

(fixed the things I didn't mention)

MLanghof:
Can't really find anything else to complain about.

> there is an option in the settings to download them: "Equipment -> Automatically download missing item images

Didn't even know that existed! :D

brather1ng:
Great!

EmmittJ:
Sorry for taking so long to get back here. I didn't come across anything
that would need to be changed. It all seems to function like you would
expect.

Next stop...accurate calculations.... haha

brather1ng:
Thanks for testing, @EmmittJ!

> Next stop...accurate calculations.... haha

Yeah, basically everything big except computation (and jewels) is updated/implemented now.

PR #457, opened 2017-03-26
Added Alternative optional replacement for float added

BlazesRus:
Requires 5 bytes instead of 4 bytes of float but stores precisely at 4 decimals places (and +- uint 16 stored in whole half).  Use AlternativeCode build to switch to replacement type usage. (Or just make new config settup with PoESkillTree_UseSmallDec_ForAttributes preprocessor flagged)


BlazesRus:
AlternativeTypes configs need fixed currently for display of the numbers on the bars still but normal non-AlternativeType builds have been tested by me to work right(Stress tested the x64 build with 300 build generations).

BlazesRus:
Compiling at x64 fixes the out of memory crash from generating too many builds with usage of lots pseudo attributes

MLanghof:
...could you elaborate on why you think this is needed?

brather1ng:
Yeah, like @MLanghof, I don't see the use in this. Floats can accurately store values with 6 significant digits, so why would we want to use something that is bigger and can store less digits?

BlazesRus:
Fixed the build now

BlazesRus:
More accurate display of the build differences

BlazesRus:
float does not store every decimal from 1-6 digits

BlazesRus:
Fixes those sometimes 0.1 to 0.2 difference in displayed attribute differences

BlazesRus:
I could add option preprocessor flag to reduce SmallDec down by 1 byte in size while still storing same amount of possible represented values and 100% accuracy stored values at small extra cost of processing time (since plenty of unused ushort value in DecimalStatus)

BlazesRus:
https://github.com/BlazesRus/PoESkillTree/releases/tag/ReleaseCompile_3_26_2017_X64
is compiled build (64 bit build)

EmmittJ:
A float has a precision, in C#, of 7 digits. If you need more then you can
use a double (16 digits) or decimal (28 digits). I am somewhat confused as
to what this implementation brings. You are just always storing up to 4
decimal places?

MLanghof:
I guess he wants to avoid the 0.99999997 kind of thing. But the solution to that is to round appropriately, not to replace floats.

WPFSKillTree/GlobalCode_ExperimentalCode/SuperDec_SmallDec_TypeConverter.cs
```diff
@@ -0,0 +1,155 @@
...
+			if (destinationType == typeof(int))
+			{
+				return (int)value;
+			}
+			return base.ConvertTo(context, culture, value, destinationType);
+		}
+	}
+}
+
+//Some Converter code from ArithmeticConverter.cs in PoESkillTree code (required license for related part of code below)
```
MLanghof: ...why are you putting license references to our code into our code?

WPFSKillTree/ViewModels/Attribute.cs
```diff
@@ -16,7 +16,7 @@ public Attribute(string text)
             Text = text;
         }
 
-        public override string ToString()
+		public override string ToString()
```
MLanghof: https://imgur.com/a/lF4AP

WPFSKillTree/GlobalCode_ExperimentalCode/SuperDec_SmallDec_TypeConverter.cs
BlazesRus: Removed

WPFSKillTree/ViewModels/Attribute.cs
BlazesRus: Replaced those with tabs with (4x Spaces)

BlazesRus:
Optional flag based config that replaces float with SmallDec for attributes values (not things like weight and position) for  a fix for the:
     * - Mana regeneration shows sometimes incorrect value (0.1 difference from in-game value).
     * - Damage type ranges shows sometimes incorrect value affecting overall DPS (1 difference from in-game value, occurs with damage conversions).
     * - Damage per Second shows sometimes incorrect value (0.1 difference from in-game value).
     * - Spell Critical Strike chance shows sometimes incorrect value (0.1 difference from in-game value).
     * - Estimated chance to Evade Attacks shows sometimes incorrect value.
listed on compute.cs (and fix for sometimes inaccurate display of hp regeneration percent display comparisons sometimes being off (sometimes displaying difference of 1 value when not really full value of difference)

PR #460, opened 2017-04-04
Gem sockets

BlazesRus:
Fixed crash on loading too many pseudoattributes and added 64 bit builds(64 bit builds fix the crash on generating too many builds at once from the any cpu build) etc


MLanghof:
This again?

Issue #465, opened 2017-05-01
Affix crafting information from the GGPK


brather1ng:
Our current Affix information is from exilemods.com. Biggest issue with that is that it's outdated, so I want to get the information directly from the GGPK (through the files I already exported in [RePoE](https://github.com/brather1ng/RePoE/tree/master/data)).

Since the GGPK's mod information is based on tags, I'll take the opportunity to replace our ItemGroup and ItemType enums by GGPK based information with ItemClass (based on ItemClasses.dat) and Tags (based on Tags.dat). This will hopefully make a lot of conversions between enums and other enums or game information much easier (e.g. it already made the ItemBaseLoader 1/3 shorter). And it should be helpful later on for the computation rewriting.

This is the last thing I want to do before starting with #463, so I thought I'd let you know.

Edit: I'll also switch over to translating mod/stat ids on-the-fly while crafting instead of using pre-translated strings. I'll need to implement translation anyway and this'll fix items with changing translations depending on the value, like Ventor's Gamble.

erickribeiro-dev:
@brather1ng will that allow the us to choose[Vaal mods ](http://pathofexile.gamepedia.com/List_of_item_corrupted_implicit_modifiers)on the items?

PR #466, opened 2017-05-21
Affix crafting information from the GGPK

brather1ng:
This is basically what I described in #465. All the old affix loading code is removed, including the Affixes.xml file, which is where most of the diff comes from.

Also, relevant RePoE files are now loaded by UpdateDB and managed by git. Those are a single line long so they can't bloat diffs. I first planned to only load them as part of the release script, but then each dev would need to manually keep their files up to date.

Turned out I had to refactor/rewrite a lot of things because of the classification enum changes (ItemClass and Tags instead of ItemGroup and ItemType) and because of doing stat translation on-the-fly, so this took a bit longer.

Most relevant changes are in `Model/Items` (and sub namespaces) and `ViewModels/Crafting`.

Known issues which I don't plan on fixing except someone things they should be fixed:

- Tier number is no longer shown below sliders as it is not that easily available anymore
- The Fishing Rod base shows more mods than it can actually have. This is caused by definining the `TwoHandWeapon` tag as `Weapon | TwoHand`, which is true for all weapons except Fishing Rod.
- If all values for a stat translation are 0, there is no translation shown above the slider as the stat has no effect and therefore no translation. I don't think any unique has 0 as the highest value for a stat, so this should rarely happen and be obvious when it happens.
- Light Radius and Accuracy shows a different translation above the accuracy slider than on the item when the slider is increased to 10. The translation on the item is correct but the one above the slider is not because it is only based on the value for that slider and that is ambigous at 10.


MLanghof:
> that is ambigous at 10.

I didn't get that part.

brather1ng:
So the mod group "LightRadiusAndAccuracy" has these mods on 1h weapons:

| | | |
| --- | --- | --- |
| of Shining | Req. Lv. 8 | 5% increased Light Radius, +(10-20) to Accuracy Rating |
| of Light |	Req. Lv. 15 | 10% increased Light Radius, +(21-40) to Accuracy Rating |
| of Radiance | Req. Lv. 30 | 15% increased Light Radius, (5-10)% increased Accuracy Rating |

If 10 is selected at the second slider, it can either be the first or the third mod. Which one it is only obvious when also taking the first value into account. The text above a slider only knows about its own value, so it shows the incorrect translation when increasing the slider to 10.

MLanghof:
I didn't even know that something like that exists in the game... I guess they handle it as three different modifiers internally but that would be more confusing here. I guess.

WPFSKillTree/Model/Items/StatTranslation/IndexHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+    public static class IndexHandlerEx
+    {
+        private static readonly IReadOnlyDictionary<IndexHandler, Func<double, double>> Handlers
+            = new Dictionary<IndexHandler, Func<double, double>>
+            {
+                { IndexHandler.Percent60OfValue, d => d * 0.6 },
+                { IndexHandler.DecisecondsToSeconds, d => d * 10 },
+                { IndexHandler.DivideBy100, d => d / 100 },
+                { IndexHandler.DivideBy100AndNegate, d => -d / 100 },
+                { IndexHandler.DivideBy100Precision2, d => Math.Round(d, 2) },
```
MLanghof: They chose some really weird names (assuming this is not a mistake on your part)...

MLanghof:
Can I just remark how insanely productive you are on this? That's 3k new handwritten LOC... It took me an hour to read through the changes alone. Massive respect!

EmmittJ:
My apologies for not getting to this. I can take a look at it tomorrow, but
I would also like to thank you. You have put so much effort and are the
reason the program is still used.

WPFSKillTree/Model/Items/StatTranslation/IndexHandler.cs
brather1ng: This is a mistake on my part.

DecisecondsToSeconds seems to be incorrect in PyPoE, where I took the handlers from. But that one is only used in an unused stat, so no way to check if the handler or the name is incorrect.

brather1ng:
Thanks @MLanghof and @EmmittJ!

> I didn't even know that something like that exists in the game... I guess they handle it as three different modifiers internally but that would be more confusing here. I guess.

Yeah, I was surprised myself when I saw it. I guess noone cares about this mod 😆 Internally they probably roll the mod first and only then check for the stats it has and roll those, which means they can't have this kind of problem.

> aside from the one negative lookbehind regex, had to look that one up

That was copied from before. Better comments probably couldn't hurt for regex stuff.

brather1ng:
> but they will need some attention if attribute rolls change?

Yeah, some of the tests verify assumptions made on the data, so they need to use the real data. Attribute rolls shouldn't change often, but if it turns out to be an issue I can add a fixed data file for testing those things.

PR #479, opened 2017-06-18
Fix build scripts with VS 2017; support C# 7

### LINKED ISSUE #476: Build scripts not working with Visual Studio 2017
Our current build scripts don't work with VS 2017 as `vsvars32.bat` [no longer exists](https://stackoverflow.com/questions/42805662/vsvars32-bat-in-visual-studio-2017). 

I talked with @EmmittJ about it and he got it working again with VS 2017, but the new used file doesn't exist on older versions. So we have two option: complicate the build scripts by adding if-cases checking for the installed VS version, or require that everyone uses VS 2017.

If everyone is already using VS 2017, requiring that would be the easier solution.

We could then also use C# 7 features. Those [should work with .NET 4.5](https://stackoverflow.com/a/42488315), meaning the program would still run everywhere it currently does.

C# 7 features: https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7 (mainly true language support for tuples and a bunch of really nice syntactic sugar)

brather1ng:
- Fix build scripts to work with VS 2017
- Update Microsoft.Net.Compilers NuGet package to support C# 7
- Update System.ValueTuple NuGet package to C# 7 version
- Use C# 7 features in one file to test compiling
- Add "Information for contributors" readme section. If someone thinks I missed something, just add it.

@EmmittJ and @MLanghof: as this wasn't as simple as I initially thought (not just changing the .bat file called in build scripts), it would be good if you can verify that both compiling/running the program and running the scripts works (with no further required setup than what I added to the readme).

The target framework is still .NET 4.5 so this shouldn't affect runnability of the compiled program, but if anyone has an old Windows or a Linux setup, it would be great if that could be confirmed.

Close #476

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,12 @@ Note: Both these options require that you can access the official Path of Exile 
 * If you find any bugs/faults please report it here on GitHub
 * Miss any features? Create an issue here or post in the forum thread
 
+## Information for contributors
+
+* The code requires Visual Studio 2017 to be compiled and run
+* It is compiled to .NET 2.5 using C# 7
+* To run the batch scripts in WPFSKillTree, your installation's Common7/Tools/ folder must be added to the PATH environment variable
+
 ## Credits
 
 * Headhorr - for his original "Unofficial Offline Skilltree Calc" http://www.pathofexile.com/forum/view-thread/19723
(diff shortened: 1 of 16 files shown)


MLanghof:
- Debug build: Successful.
- Release build: Successful after adding Common7/Tools to PATH.

- build-locale.bat: Successful.
- dist-clean.bat: Successful (as in, no error the one time I ran it).
- dist-release.bat: Fails due to unit tests. Everything before that is successful from what I can tell.
- dist-update.bat: Successful.

brather1ng:
@EmmittJ would be great if you could test that it's all still working for you

EmmittJ:
These are all changes I made locally (except adding Tuples), and it worked for the last release. I can see if it works later today. I am about to head into work, sorry.

brather1ng:
Didn't know what you did locally was equivalent to this. Fine for me then. Probably best to verify I didn't break anything for you at some point before the next release, but that can wait.

EmmittJ:
Yep, I will double check, but I had installed 2017 and had to change the build stuff for the last release. I didn't want to push it until I had that talk with you on discord

PR #493, opened 2017-08-19
Extended v2

BlazesRus:
All 21 Jewel Slots added to ItemView
Minimum Value option added to skilltree generator for certain configs (with alternative csvs score calculation for those configs)


BlazesRus:
Still need to finish adding pseudo-attribute display part later. 
Use ExtendedGUI config to enable MinimumValue settings (and alternative score calculations).

BlazesRus:
Which naming scheme should I use for release builds?

BlazesRus:
Ok I'll create version of branch without the optional aliases and multiple configs etc(Just the Jewel slots and 64 bit compile option).

PR #497, opened 2017-10-05
fixed it so my and probably other account names fit

ImaginaryDevelopment:
my account name wasn't allowed in these dialog boxes. as it is 30 characters

### DIFF
--- a/WPFSKillTree/Views/Builds/EditBuildWindow.xaml
+++ b/WPFSKillTree/Views/Builds/EditBuildWindow.xaml
@@ -46,7 +46,7 @@
                 <Label Margin="4,4,4,2">
                     <l:Catalog Message="Account name"/>
                 </Label>
-                <TextBox Height="23" Margin="4,2,4,4" MaxLength="24"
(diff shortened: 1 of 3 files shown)


brather1ng:
Whoops, no idea why we limited it to 24 characters. Thanks for the PR!

I checked PoE's account creation page and it looks like account names are allowed to be 31 characters long. Would make sense to change our limit to 31 too, so we don't run into this problem again.

ImaginaryDevelopment:
Ok, I guess you found an easier way, I'm having some code search the poe stash api for the longest account name.
Would you be up for/ok with a submission that added or migrated some code to an F# assembly/project in the solution?

I'll update the files to 31

brather1ng:
I never did anything with F# and a mixed language project would make contributing and maintaining more difficult, so I'd rather let it stay C#-only. But that's just my opinion. @EmmittJ and @MLanghof, do you have experience with F#?

MLanghof:
None.

EmmittJ:
My experience is very limited in F#, so I wouldn't say I would be able to support any if it was put into the project. Are there things would directly benefit from assembly? I don't think our process times are very high.

PR #500, opened 2018-01-01
Computation Rewriting Part 1: Data and Parsing

brather1ng:
Took a while because I didn't do much after the first month, but here we go!

This contains both the data and the logic for parsing stat translations into object representations, though the data is not complete at all (~94% of unique skill tree stats are successfully parsed, I didn't specifically add any item stats except some jewels so the percentage will be much lower for items).

The interfaces for the object representations only have a stub implementation that basically just builds a string representation again, which is the parsing output seen below.

This isn't integrated into the main application in any way (which means merging this PR does not add features, it doesn't matter if it takes a while to get reviewed). You can start the console program in the `PoESkillTree.Computation.Console` project to manually enter stats to be parsed (see example output below).

`PoESkillTree.Computation.Console` has a benchmark mode which parses all skill tree stats and times that. That benchmark (see the output below) shows that parsing basically takes no time at all: under 0.5 ms for a not yet seen stat line after the first few parses. I do find that quite surprising with all the regex stuff happening.

## Quick overview over the code

I've tried doing this properly with Test-driven development and Dependency Injection (without DI library), so the code looks quite different from what I've written previously. This was my first time doing that but it does have a lot more unit tests and should be a lot more maintainable.

The code is split into multiple projects (all prefixed with `PoESkillTree`).

- `Common`: Just some Enums moved over from the main project and a few utility functions.
- `Computation.Common`: Contains the interfaces and data structures implemented/used by the Data and Parsing.
- `Computation.Common.Tests`: Contains the unit tests for `Computation.Common`.
- `Computation.Parsing`: Contains the parsing logic.
- `Computation.Parsing.Tests`: Contains the unit tests for `Computation.Parsing`.
- `Computation.Data`: Contains the actual data. The root namespace is the most important thing here. It is were most changes for adding support for new stats should happen (everything that doesn't require new mechanics).  
  I tried to design the `Computation.Common.Builders` interfaces in a way that makes the matchers implementations as readable and adding new stats as simple as possible.
- `Computation.Data.Tests`: Contains the unit tests for `Computation.Data`.
- `Computation.Console`: Contains stub implementations of `Computation.Parsing.Builders` (which are not really interesting) and the console program explained above.
- `Computation.IntegrationTests`: Contains tests that verify that the data follows the rules (e.g. no illegal references), tests that verify all stats from a file are parsed successfully unless explicitly added to a file of unsupported stats and tests that check the correctness of the full parsing output for a few stats.

This is a lot of code (but still a lot less logic than it looks from the file/line counts), I'll try to develop the next computation parts in smaller Pull Requests. However, there were some major refactorings towards the end, so I don't think this could have been split up.

## How does the parsing work?

What follows is a simplified explanation. For the full picture, it's best to look at the source. `ParsingData`  in `Computation.Data` and `Parser` in `Computation.Parsing`, being the composition roots of the respective projects, are good starting points.

Every `IStatMatchers` implementation in the `Computation.Data` project represents on step in parsing a stat line. The parser runs through these like a state machine (see `Computation.Data.Steps.Stepper`). The implementations contains tuples of regexes and the modifier parts that are added to the parsing output if the regex matches. In one step, only the tuple whose regex matches the longest substring of the remaining (unparsed) stat is taken. 

The regex patterns are not taken directly. They are expanded before being matched. Two things are expanded:

- Value placeholders ("#") are replaced by a pattern matching numbers. This means that a) you don't have to repeat number patterns everywhere and b) that the values can be referenced from the modifier building part.
- Reference placeholders can be used to reference other `IStatMatchers` implementations and the `IReferencedMatchers` implementations. These will be replaced by all (recursively expanded) regex patterns of the referenced implementation and can be referenced from the modifier building part.

Expanding and resolving in a generic way like this does make the parsing logic a lot more complicated, but I think it is worth it as it allows very flexible matcher implementations and saves a lot of duplication.

## Technical notes

The new test projects are written using NUnit instead of Microsoft's testing library, basically because it is much better, e.g. it allows parameterizing tests with attributes alone. This leads to tests that are not detected by Visual Studio by default. I have included the NUnit test adapter NuGet package that claims to add support to Visual Studio, but I'm using ReSharper, which detects them by default, so I don't know if and how well that works.  
The new test projects are executed as part of the Test target of the build script.

The projects are now using .Net 4.5.2 instead of .Net 4.5, which hopefully doesn't affect users. See https://github.com/PoESkillTree/PoESkillTree/commit/979da34d2ffb3cf01c1e0c2bc8a29c4bec0d718d for more information.

The projects are using the latest minor C# version (7.2) instead of the latest major version (7, VS's default value) because 7.1 has some improvements related to tuples and the language version doesn't change the compiled binary anyway.

---

Example output of `PoESkillTree.Computation.Console`:

```
Enter a stat line to be parsed (or 'benchmark' to time stat parsing)
> Gain 30 Mana per Grand Spectrum
Stat: Socketed Grand Spectrum jewels
  Form: Base add
  Value: 1
  Condition: unconditional
Stat: Mana
  Form: Base add
  Value: (30 * Value of Socketed Grand Spectrum jewels)
  Condition: unconditional
> 10% increased fire damage
Stat: Fire Damage
  Form: Percent increase
  Value: 10
  Condition: unconditional
> Can have up to 1 additional Totem summoned at a time
Stat: Maximum Skills.Where(has keywords [Totem]) combined instance count
  Form: Base add
  Value: 1
  Condition: unconditional
> Auras you Cast grant 3% increased Attack and Cast Speed to you and Allies
Stat: Skills attack/cast speed added to skills Skills.Where(has keywords [Aura])
  Form: Percent increase
  Value: 3
  Condition: unconditional
> Your Offering Skills also affect you
Stat: apply stats of [Bone Offering, Flesh Offering, Spirit Offering] to Self
  Form: Total override
  Value: 1
  Condition: unconditional
> Every 10 seconds, gain 100% increased Elemental Damage for 4 seconds
Stat: Buff rotation for 10 seconds { Fire, Lightning, Cold Damage as Buff for 4 seconds }
  Form: Percent increase
  Value: 100
  Condition: unconditional
> For each Element you've been hit by Damage of Recently, 8% reduced Damage taken of that Element
Stat: Fire Damage taken
  Form: Percent reduce
  Value: 8
  Condition: If any Hit (with Fire damage) recently by Any Entity against Self
Stat: Lightning Damage taken
  Form: Percent reduce
  Value: 8
  Condition: If any Hit (with Lightning damage) recently by Any Entity against Self
Stat: Cold Damage taken
  Form: Percent reduce
  Value: 8
  Condition: If any Hit (with Cold damage) recently by Any Entity against Self
>
```

Benchmark output:

```
Initialization (parsing 1 made-up stat):
  228 ms
Batch | Total ms | ms/line | ms/distinct | Batch size | #Distinct
=================================================================
    1 |      171 |    0,58 |        0,73 |        294 |       234
    2 |       68 |    0,23 |        0,38 |        294 |       180
    3 |       62 |    0,21 |        0,40 |        294 |       156
    4 |       53 |    0,18 |        0,38 |        294 |       139
    5 |       37 |    0,13 |        0,34 |        294 |       110
    6 |       45 |    0,15 |        0,42 |        294 |       108
    7 |       50 |    0,17 |        0,51 |        294 |        98
    8 |       28 |    0,10 |        0,31 |        294 |        89
    9 |       31 |    0,11 |        0,39 |        294 |        80
   10 |       26 |    0,09 |        0,36 |        285 |        73
Total |      571 |    0,19 |        0,45 |       2931 |      1267
Successfully parsed lines:
  2825/2931 (96,4%)
Successfully parsed distinct lines:
  1196/1267 (94,4%)
```


brather1ng:
So, does anyone plan to look over this? I don't mind much if not, but I then can merge this. I completely understand if this is too much to look over.

Still, I would greatly appreciate it if someone could at least look over the .Data project's root namespace. It's less about whether every single line of matcher data is correct (it probably isn't), but more about feedback on how understandable and discoverable adding to/maintaining the data is. Looking over the .Common (the builder interfaces are what's responsible for the last sentence) and .Parsing projects would also be appreciated, everything else doesn't matter that much.

PR #504, opened 2018-03-01
Use stronger cryptographic standards

### LINKED ISSUE #503: Check for Updates fails with SSL error
"The request was aborted: Could not create SSL/TLS secure channel."

Seems like GitHub changed another thing in their API

Nhowka:
Fixes #503


EmmittJ:
Looks good. Thanks

PR #539, opened 2019-12-05
Move to .NET Core 3 & C# 8

brather1ng:
This PR does two major things:

- All projects now target .NET Core 3.1 instead of .NET Framework. .NET Framework isn't really extended anymore and it allows full C# 8 compatibility.
- Nullable reference types were enabled in all projects.

None of those change anything functionally, but I used both to do a lot of clean up:

- All loose DLLs required by the program itself were replaced by NuGet packages.
- Removed the ItemFilter feature. It wasn't updated for a while and I couldn't be bothered to migrate it.
- Improved release.xml and related things. It now uses the dotnet command where possible. It also allowed me to use its AssemblyInfo and AppManifest generation instead of our custom logic. Now `WPFSKillTree.csproj` has to be adjusted when doing releases as there is no longer a `Version.resx`. Also, there's no longer a GUID that has to be generated. The Product name has to be unique anyway.
- While having to touch most files for the nullability annotations anyway, I removed a lot of dead code.

Because most users probably don't have the required .NET Core runtimes installed, publishing is done as a self-contained exe. That increases the artifact size by ~50 MB, but avoids all kinds of .NET installation problems.

The next release should be a new product release. I don't know whether updating an installation from a previous Framework version to a Core version would work properly. But with Conquerors of the Atlas coming up and this being a 3.0 alpha, it would make sense even without these changes.


EmmittJ:
A lot of files here, but overall it looks alright. Make sure you end up updating the wiki page https://github.com/PoESkillTree/PoESkillTree/wiki/How-To:-Release. I like the move to .NET Core 3.1

Everything appears to run/build in Visual Studio 16.4 as well (required for .NET Core 3.1)

brather1ng:
The file amount is mostly caused by having to touch a lot of them for nullability annotations.

> Make sure you end up updating the wiki page https://github.com/PoESkillTree/PoESkillTree/wiki/How-To:-Release.

Done, I hope it's clear enough

EmmittJ:
Yeah, I figured. Nullable is really nice though, so this is a great improvement. I think they are clear enough! Thanks

PR #541, opened 2019-12-11
import updates done

myblindy:
Made a few small changes to allow importing items from an account name, providing the list of characters on the account in a dropdown for easy selection.

And maybe I missed it, but as far as I can tell it was only importing items (and a separate importer for stashes), I added the ability to import passive skill points to the items importer for free!

### DIFF
--- a/WPFSKillTree/ViewModels/Equipment/DownloadItemsViewModel.cs
+++ b/WPFSKillTree/ViewModels/Equipment/DownloadItemsViewModel.cs
@@ -1,6 +1,12 @@
-﻿using System.ComponentModel;
+﻿using System;
+using System.Collections.ObjectModel;
+using System.ComponentModel;
 using System.Diagnostics;
+using System.Linq;
+using System.Net.Http;
 using System.Windows.Input;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using PoESkillTree.Common.ViewModels;
 using PoESkillTree.Localization;
 using PoESkillTree.Model.Builds;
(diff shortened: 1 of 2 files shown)


EmmittJ:
You should use the passive tree parser already in the program instead of writing your own there. Also, hard coding the 3.9 in the URL isn't the best way to generate urls since you could be loading a very old tree using the character import. 

 I haven't tried this, but what happens when you try to use this feature offline? The way it is currently is so that the program is mainly offline compatible with a few instances of requiring an internet connection (the user is prompted)

myblindy:
Was that the reason why you had to copy links back and forth, to allow the tool itself to work offline? That’s fine if so, it’s just not the tool for me — I’m online both when I play POE and when I use POE tools.

EmmittJ:
Yeah, basically. We do have features that require internet (like updating the skill tree, updating the application, etc.), but they warn the user that an internet connection is required before doing the action. It would still be nice to support the offline aspect of this window. 

There are a still a few issues with your commit. You should put your account model in the correct models folder. You should be using the skill tree deserializer as to not repeat the code used there.

brather1ng:
You now always import both items and passive tree. That's not a good idea, imo, because it e.g. overwrites your planned tree if you just want to update your items. Tree and item import should be two separate things.

Getting the item data directly through HTTP calls doesn't work if your characters aren't public. Your solution completely removes the current functionality for non-public characters. I'm not sure how querying an account's characters works, but if it's similar, that also doesn't work if the character page is not set to public.

The select character was previously loaded from and save to the build, meaning you only have to enter your character once. If I'm reading it correctly, that functionality is also gone with your changes.

PR #549, opened 2019-12-30
Add Ribbon Control + some general UI improvements

### LINKED ISSUE #547: Organize buttons from different UI elements into a single one
Currently, our functionality reachable through buttons is placed all over the place. Some is in the Saved Builds expander, some is in the Equipment tab and some doesn't even have buttons (Skill Tree Generator). This would get even worse with #543 and #544.

To improve discoverability of features, these buttons should all be organized into a single UI element that's always visible. The obvious solution here is either a simple icon bar or ribbon control, but I'm open to other ideas.

New Top Level Build/Folder are exceptions imo, they could just be moved to the top.

brather1ng:
Not gonna write much, the screen shots are much more informational anyway.

image
image

If you notice anything that can be improved from the above images or testing it yourself, I'm open.

Closes #547


EmmittJ:
I think the Passive Tree/Equipment/Computation "Tabs" should be moved to a more permanent place. Having the flyouts underneath that center bar would be nice. That way the only thing that changes is view window for the passive tree. 

I think it would be nice to have the passive points always visible as well (back on the center bar). They kind of hover awkwardly on the skill tree. 

Does the Summary of changes button actually do anything? It seems the Tree comparison button is the only thing that actually does something. 

Undo/Redo buttons aren't faded when there is nothing to undo/redo. 

I do like the ribbon though, it is very nice.

SpaceOgre:
I like the look of the ribbon, but it makes the other UI elements look a bit cramped in there.

Here are some suggestions:

* The regex checkbox in build summery needs some padding both left and right.
* The row of controls under the skill tree could be a bit smaller and thus creating more space.

brather1ng:
Thanks for the responses.

> I think the Passive Tree/Equipment/Computation "Tabs" should be moved to a more permanent place. Having the flyouts underneath that center bar would be nice. That way the only thing that changes is view window for the passive tree.

@EmmittJ I'm not sure about this. Currently, the tab headers are directly above what they are changing. Moving them to the left above the build summary panel (if it's active) would de-attach them too much from their tabs again, imo. Also, the bandit/level/class controls will probably move into the build summary at some point when that's redone, so that panel needs all the vertical space it can get.

> Does the Summary of changes button actually do anything? It seems the Tree comparison button is the only thing that actually does something.

It does. It sums up the stats of the nodes you gain/lose when hovering a node. This will be replaced by something connected to computation later (preview of the impact on the most important computed stats).

> The row of controls under the skill tree could be a bit smaller and thus creating more space.

@SpaceOgre In what way could it be smaller? Reducing the height a bit would be possible, but would require changing the padding and/or the font size of the controls. Reducing the width wouldn't create much space because it'd need to be either floating or split into left and right aligned controls.

> Not sure how I feel about the border around the tree, on the images I get the feeling that it feels more cramped since the border is close to other elements. But I have not tested the program yet so it might feel different then. But maybe add a border to the bottom of the top center bar and a border to the top of the bottom bar instead.

I like it a lot more than the black background and faded edges we had before. Though it could probably be a lot smaller and still work as a separation or might not be required at all. Zoomed in, you can't see a border and it still looks fine, imo.

I do agree on everything else.

SpaceOgre:
> >The row of controls under the skill tree could be a bit smaller and thus creating more space.

>@SpaceOgre In what way could it be smaller? Reducing the height a bit would be possible, but would require changing the padding and/or the font size of the controls. Reducing the width wouldn't create much space because it'd need to be either floating or split into left and right aligned controls.

I was thinking that reducing the width ever so slightly might do the trick. So reduce the skill tree url input with like 10px and just evenly distribute the extra space between the controls to get a bit more "air" in.

EmmittJ:
My only major problem with the tab headers is that they change position and can be blocked off by the other elements in the center bar there since the flyout shrinks the horizontal space by a lot. 

Visual Studio uses a toggle button for their search boxes which would reduce the space in both places pretty significantly.

MediatR could simplify a lot of the skill tree related stuff. The skill/unskill/hover on a node is very highly linked to the zoom border.

EmmittJ:
Hit close in comment on mobile by accident, reopened

brather1ng:
Should've gotten everything. Updated screen shots:

image
image

> My only major problem with the tab headers is that they change position and can be blocked off by the other elements in the center bar there since the flyout shrinks the horizontal space by a lot.

Changing position is the expected behavior with these kinds of side panels, imo. E.g. in Visual Studio the tab headers move with the tab contents when you dock something to the left side.

I agree on your second part. Those other elements in the center aren't meant to stay and will (at least partially) move into the build summary later on. Though I doubt anyone would use the program at a width were it becomes problematic while having both panels open. The equipment and computation tabs aren't really usable then either.

SpaceOgre:
Looks very good!

MLanghof:
Very impressed how this is all coming together. Looks really good!

PR #553, opened 2020-01-15
New "Skills" tab

### LINKED ISSUE #552: New "Skills" tab as an alternative to the "Edit Socketed Gems" dialog on items
(continuation of the [comments](https://github.com/PoESkillTree/PoESkillTree/issues/547#issuecomment-568307288) in #547)

My suggestion:

- Add a new tab "Skills".
- It contains one tab per item slot.
- The slot tab's contents are the same as the current "Edit Socketed Gems" dialog (to keep it simple).
- The tab headers preview the active skills for that slots.
- Tab header tooltips are a more detailed preview.
- Maybe add tooltips to the gems themselves. The Engine already provides tooltip information as part of skill definitions.

brather1ng:
I basically did what I wrote in #552.

Screenshots of how it looks:
image
image

The "Equipped item's maximum number of sockets: n" bit at the bottom looks a bit lost with reasonable numbers of skills, but the skill list needs to be stretched for scroll bars and dialog resizing to work properly.

Closes #552


EmmittJ:
Yeah, this seems much more intuitive than it being coupled with equipment. I am not sure you actually need the equipment item's maximum sockets since most people don't care too much. Gem swaps, different options for socket colors, etc. end up making you go over that number anyway.

brather1ng:
I agree that it's not very useful. But it's there, so might as well keep it.

PR #559, opened 2020-03-19
Passive skill tree update

EmmittJ:
The update utilizes PoESkillTree.Engine Passive Skill Deserialize. Most of the file changes are due to the SkillNode -> PassiveNodeViewModel. The main files to look at are the PassiveTreeViewModel, PassiveNodeGroupViewModel, and PassiveNodeViewModel. 

This also includes all the updates to the drawing code to draw node/backgrounds in the correct locations, and some improvements around clicking items on the tree.


WPFSKillTree/ViewModels/PassiveTree/PassiveNodeViewModel.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        public PassiveNodeViewModel() : this(new JsonPassiveNode()) { }
+        public PassiveNodeViewModel(JsonPassiveNode jsonPassiveNode, PassiveNodeGroupViewModel? group = null)
+        {
+            JsonPassiveNode = jsonPassiveNode;
+            PassiveNodeGroup = group;
+            InitializeAttributes();
+        }
+        public PassiveNodeDefinition PassiveNodeDefinition => PassiveNodeDefinition.Convert(JsonPassiveNode);
+        public CharacterClass? StartingCharacterClass { get => JsonPassiveNode.StartingCharacterClass; set => JsonPassiveNode.StartingCharacterClass = value; }
```
brather1ng: Of all these setters, I found two that are used: StatDescriptions and Id. StatDescriptions can be made private and Id is only used in tests, which are also the only uses of the parameterless test, so it can be a parameter there.

All other setters can be removed. Those properties aren't really something that should be changed anytime.

PassiveNodeGroupViewModel and PassiveTreeViewModel also have a few unused setters.

WPFSKillTree/ViewModels/PassiveTree/PassiveNodeViewModel.cs
```diff
@@ -0,0 +1,147 @@
...
+        public string[]? Recipe { get => JsonPassiveNode.Recipe; set => JsonPassiveNode.Recipe = value; }
+        public JsonExpansionJewelSocket? ExpansionJewelSocket { get => JsonPassiveNode.ExpansionJewelSocket; set => JsonPassiveNode.ExpansionJewelSocket = value; }
+        public HashSet<ushort> OutPassiveNodeIds => JsonPassiveNode.OutPassiveNodeIds;
+        public HashSet<ushort> InPassiveNodeIds => JsonPassiveNode.InPassiveNodeIds;
+        public bool IsSkilled { get => JsonPassiveNode.IsSkilled; set => JsonPassiveNode.IsSkilled = value; }
```
brather1ng: There are multiple unused properties in here that I don't really mind, but this one doesn't really make sense for what this class is used for and might even be misleading.

WPFSKillTree/TreeDrawing/JewelDrawer.cs
```diff
@@ -91,12 +92,8 @@ public void Draw()
         private void Draw(DrawingContext drawingContext, ushort nodeId, JewelType jewelType)
         {
             var node = _skillNodes[nodeId];
-            var (size, brush) = _brushes.GetOrAdd((jewelType, node.ExpansionJewel != null), CreateBrush);
-            drawingContext.DrawRectangle(brush, null,
-                new Rect(node.Position.X - size.Width,
-                    node.Position.Y - size.Height,
-                    size.Width * 2,
-                    size.Height * 2));
+            var (size, brush) = _brushes.GetOrAdd((jewelType, node.PassiveNodeType != PassiveNodeType.ExpansionJewelSocket), CreateBrush);
```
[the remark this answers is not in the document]
EmmittJ: This has been resolved.

WPFSKillTree/TreeDrawing/JewelDrawer.cs
```diff
@@ -91,12 +92,8 @@ public void Draw()
...
+            drawingContext.DrawRectangle(brush, null, new Rect(node.Position.X - size.Width, node.Position.Y - size.Height, size.Width, size.Height));
```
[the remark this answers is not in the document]
EmmittJ: This has been fixed be shifting the position rather than drawing at twice the height/width

WPFSKillTree/ViewModels/PassiveTree/PassiveNodeViewModel.cs
EmmittJ: These have been cleaned up

WPFSKillTree/ViewModels/PassiveTree/PassiveNodeViewModel.cs
EmmittJ: ExpansionJewelSocket will be needed if we plan to support that piece. I have removed the IsSkilled property

brather1ng:
This really did clean up a lot of stuff, looks good

